PR #41, opened 2015-03-16
Feature/export modes handling#11

wasabideveloper:


das-g:
against master instead of against develop
please create a new request

PR #47, opened 2015-03-20
Feature/export data download#26

wasabideveloper:


source/excerptexport/views.py
```diff
@@ -65,6 +75,35 @@ def create_excerpt_export(request):
...
+
+def download_file(request):
+
+    file_id = int(request.GET['file'])
+    output_file = get_object_or_404(OutputFile, public_identifier=file_id, deleted_on_filesystem=False)
+    if not output_file.file:
+        return HttpResponseNotFound('<p>No output file attached to output file record.</p>')
+
+    download_file_name = settings.APPLICATION_SETTINGS['download_file_name'] % {'id': output_file.public_identifier, 'name': os.path.basename(output_file.file.name)}
+    absolute_file_path = os.path.abspath(settings.APPLICATION_SETTINGS['data_directory'] + '/' + os.path.basename(output_file.file.name))
```
das-g: Why are we only using the `basename`, here?

das-g: `settings.APPLICATION_SETTINGS['data_directory']` is already absolute. Give reason why it has to be processed by `abspath` again, or just use it as-is.

PR #50, opened 2015-03-23
Feature/persist export requests #21

das-g:
Implements #21, but doesn't yet handle predefined country/region (administrative area) excerpts

### DIFF
--- a/source/excerptexport/views.py
+++ b/source/excerptexport/views.py
@@ -6,6 +6,7 @@ from django.template import RequestContext, loader
 from django.contrib.auth.decorators import login_required
 from django.contrib.auth.models import User
 
+from excerptexport.models import ExtractionOrder
 from excerptexport.models import Excerpt
 from excerptexport.models import BoundingGeometry
 from excerptexport import settings
@@ -40,6 +41,10 @@ def create_excerpt_export(request):
     if request.POST['form-mode'] == 'existing_excerpt':
         existingExcerptID = request.POST['existing_excerpt.id']
         viewContext = { 'use_existing': True, 'excerpt': existingExcerptID }
+        ExtractionOrder.objects.create(
+            excerpt_id = existingExcerptID,
+            orderer = request.user
+        )
 
     if request.POST['form-mode'] == 'create_new_excerpt':
         excerpt = Excerpt(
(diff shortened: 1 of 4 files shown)


source/excerptexport/views.py
```diff
@@ -36,10 +37,13 @@ def new_excerpt_export(request):
 
 @login_required(login_url='/admin/')
 def create_excerpt_export(request):
-    viewContext = {}
```
wasabideveloper: Maybe it would improve the readability and traceability where the variable is created?

source/excerptexport/views.py
```diff
@@ -59,8 +63,13 @@ def create_excerpt_export(request):
...
         bounding_geometry.save()
 
-        viewContext = { 'use_existing': False, 'excerpt': excerpt, 'bounding_geometry': bounding_geometry }
+        viewContext = { 'excerpt': excerpt, 'bounding_geometry': bounding_geometry }
+        ExtractionOrder.objects.create(
+            excerpt = excerpt,
+            orderer = request.user
+        )
 
+    viewContext['use_existing'] = 'existingExcerptID' in vars() # TODO: The view should not have to know
```
wasabideveloper: Maybe a better solution: set 'excerpt' OR 'existing_excerpt' and let the view decide by check what is defined?

source/excerptexport/views.py
```diff
@@ -36,10 +37,13 @@ def new_excerpt_export(request):
 
 @login_required(login_url='/admin/')
 def create_excerpt_export(request):
-    viewContext = {}
     if request.POST['form-mode'] == 'existing_excerpt':
         existingExcerptID = request.POST['existing_excerpt.id']
-        viewContext = { 'use_existing': True, 'excerpt': existingExcerptID }
+        viewContext = { 'excerpt': existingExcerptID }
```
wasabideveloper: TODO: refactor to snake_case. Overlooked by myself earlier :-(

das-g: @wasabideveloper we should get flake8 up an running, anyway (if possible as a git pre-commit hook). That should force us into proper naming habits and a standardized code style.

source/excerptexport/views.py
das-g: The current `viewContext` (the functionality of which I tried to preserve for now) is only useful for development/debugging anyway. The `viewContext` we finally want for `create_excerpt_export` will be creatable completely from the finished `ExtractionOrder`, so can be filled in at just one place in the code instead of scattered around like now.

In fact, this POST-request specific view shouldn't build a `viewContext` at all. Instead, it should redirect to a GET of a separate view of the newly created object. (See https://docs.djangoproject.com/en/1.7/intro/tutorial04/, first `polls/views.py` snipped and last bullet points of the 'rundown' following it)

source/excerptexport/views.py
das-g: That is in fact what I'm doing here.

PR #53, opened 2015-03-24
Feature/trigger data conversion service#24

wasabideveloper:


source/excerptexport/models/extraction_order.py
```diff
@@ -14,6 +14,7 @@ class ExtractionOrderState(enum.Enum):
     CANCELED = 5
 
 
+
```
das-g: Aren't two blank lines enough?

source/excerptexport/models/extraction_order.py
```diff
@@ -22,5 +23,11 @@ class ExtractionOrder(models.Model):
     orderer = models.ForeignKey(User, related_name='extraction_orders')
     excerpt = models.ForeignKey(Excerpt, related_name='extraction_orders')
 
+    def get_number_of_output_files(self):
+        # prevent cyclic import problems
+        from .output_file import OutputFile
+        return OutputFile.objects.filter(extraction_order=self).count()
```
das-g: If I've understood [following relationships "backwards"](https://docs.djangoproject.com/en/1.7/topics/db/queries/#backwards-related-objects) correctly,
```python
return output_files.count()
```
should do the same. (If you hadn't set `related_name='output_files'` [below](https://github.com/geometalab/osmaxx/pull/53/files#diff-113aee21f93c49ee542d3ee53079e12dR18), it'd be
```python
return output_file_set.count()
```
or similar.)

This should make the `import` unnecessary, too. :-)

PR #64, opened 2015-04-07
Feature/login

njordan-hsr:
* Login with possibility for OpenID and such
* Separate Login Screen
* Refactoring of the layout (templates) for easier overriding and integration


source/excerptexport/templates/excerptexport/base.html
```diff
@@ -0,0 +1,8 @@
+{% extends 'base.html' %}{% load static %}
+
+{% block extra_head %}
+    <link rel='stylesheet' type='text/css' href='{% static "excerptexport/libraries/leaflet-locationfilter/src/locationfilter.css" %}' />
+{% endblock %}
+
+{% block main %}
+{% endblock %}
```
das-g: Please end all non-empty text files with a newline.

njordan-hsr: Thank you. Should be corrected now.

das-g:
In the future, please try to make your commits semantically atomic:
* The changes to `requirements.txt` should have been in the same commit as the feature implementation that make the new requirements necessary.
* The actual feature implementation (introducing the login widget and getting it to work) could have been in a separate commit from the refactorings (file movements and path literal changes to reference these files at their new locations, maybe even with the simplification of the views with higher-level helper functions in yet another commit). This would have made review a bit easier.

(Don't bother to re-arrange this request's code changes like that. This is just a suggestion for subsequent changes.)

das-g:
Tried code on this branch and noticed the following issues:
* http://127.0.0.1:8000/excerptexport/ always shows the "Sign In"/"Sign Up" buttons, even when already logged in and not yet logged out again. (One can tell whether one is logged in by clicking the 'New excerpt export' link and see whether the form shows up.)
* After Logging in with user name and password, one is always redirected to the **non-existant** http://127.0.0.1:8000/accounts/profile/ , even when the login form page was invoked with a 'next' parameter (e.g. http://127.0.0.1:8000/accounts/login/?next=/excerptexport/new/)
* If I've understood Stefan correctly, we don't want to store users' passwords, thus why we just work by whitelisting OpenID identifiers. So we should not have any classic "Sign Up" capability. (We might later offer an online way to request an OpenID to be whitelisted, but that is low-prio. Until then, we'll just whitelist them manually by offline request.)

njordan-hsr:
Can you reject this pull request, so I can fix the issues mentioned? Or is it better to just leave it open?

das-g:
I'm OK with just leaving it open. It will be auto-updated with whatever to push to your branch. If you'd like to close the request without merging (there is no 'reject' on GitHub pull requests, I think), you may do so yourself by clicking "Close and comment" below the comment box on this request page's "Conversation" tab.

Also feel free to declare any of the identified issues out-of-scope of your request and to file issues for them on https://github.com/geometalab/osmaxx/issues to handle them later / have them handled by someone else of us.

source/osmaxx/templates/base.html
```diff
@@ -0,0 +1,57 @@
+<!DOCTYPE html>
```
wasabideveloper: In excerptexport app we used layouts (e.g. source/excerptexport/templates/excerptexport/layouts/base.html) for top level templates. This file looks like a top level template. Maybe you should move it to a layout directory.

source/osmaxx/templates/base.html
```diff
@@ -0,0 +1,57 @@
...
+    <link rel="stylesheet" href="http://cdn.leafletjs.com/leaflet-0.7.3/leaflet.css" />
+    <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.1/css/bootstrap.min.css" />
+    <link rel="stylesheet" type="text/css" media="screen" href="https://silviomoreto.github.io/bootstrap-select/bootstrap-select.min.css" />
+    {% block extra_head %}{% endblock %}
+    <link rel='stylesheet' type='text/css' href='{% static "excerptexport/stylesheets/main.css" %}' />
+</head>
+<body>
+{% block body %}
+
+    {% if messages %}
```
wasabideveloper: Maybe it makes sense to move this account specific template parts before header to a partial and use an include?

source/excerptexport/templates/excerptexport/create_excerpt_export.html
```diff
@@ -1,4 +1,4 @@
-{% extends "excerptexport/layouts/base.html" %}
```
wasabideveloper: What's the reason for flatting the template structure? We used layouts, templates and partials to structure the different types of templates.

das-g: I guess the flat structure is needed for the more toplevel helper functions to work. Is that the case, @njordan-hsr ?

PR #67, opened 2015-04-07
Feature/deployment container configuration

wasabideveloper:


deployment/README.md
```diff
@@ -0,0 +1,77 @@
...
+2. Build image:
+
+    ```shell
+    docker build --tag=osmaxx --no-cache=true deployment/
+    ```
+3. Run container:
+
+    ```shell
+    # map host port 8080 to port 80 of container
+    docker run -d -p 8080:80 osmaxx /bin/bash -c "service postgresql start && /usr/sbin/apache2ctl -D FOREGROUND"
```
das-g: Could `service postgresql start && /usr/sbin/apache2ctl -D FOREGROUND` (or equivalent) be put into a script within the container, so that one could just

    docker run -d -p 8080:80 osmaxx /path/to/start-server
?

deployment/Dockerfile
```diff
@@ -0,0 +1,76 @@
...
+USER postgres
+    # Fix pg template ascii bug
+    # @source: http://stackoverflow.com/questions/16736891/pgerror-error-new-encoding-utf8-is-incompatible
+    RUN service postgresql start &&\
+        psql -c "UPDATE pg_database SET datistemplate = FALSE WHERE datname = 'template1';" &&\
+        psql -c "DROP DATABASE template1;" &&\
+        psql -c "CREATE DATABASE template1 WITH TEMPLATE = template0 ENCODING = 'UNICODE';" &&\
+        psql -c "UPDATE pg_database SET datistemplate = TRUE WHERE datname = 'template1';" &&\
+        psql -c "\c template1" &&\
+        psql -c "VACUUM FREEZE;"
```
das-g: It would probably make sense to have little (p)sql script files that we reference here, instead having SQL commands inline in the `Dockerfile`.

wasabideveloper: Docker runs only changed commands if you rebuild a container. If you move it to a script, docker can not detect changes. 
@das-g If you think it's not a problem, we can try to move it (Needs testing again).

deployment/README.md
wasabideveloper: I tried but it didn't work and I had not enough time to play around with. I think it's possible but it will need some time to get it working.

Alternative way to keep the container running with a background process:
```shell
CMD /path/to/serverStart.sh && /bin/bash
```

deployment/Dockerfile
das-g: Ah, I wasn't aware of that. In that case, let's leave it in the `Dockerfile`.

deployment/Dockerfile
```diff
@@ -0,0 +1,79 @@
...
+ADD source.tar /var/www/eda/
+COPY requirementsProduction.txt /var/www/eda/requirementsProduction.txt
+RUN mv /var/www/eda/source /var/www/eda/projects;
+
+RUN cd "/var/www/eda/projects"
+RUN /var/www/eda/environment/bin/pip3 install -r /var/www/eda/requirementsProduction.txt
+RUN service postgresql start &&\
+    /var/www/eda/environment/bin/python /var/www/eda/projects/manage.py migrate
+RUN service postgresql start &&\
+    echo "from django.contrib.auth.models import User; User.objects.create_superuser('admin', '', 'osmaxx')" | /var/www/eda/environment/bin/python /var/www/eda/projects/manage.py shell
```
njordan-hsr: I'm not sure if it is a good idea to have a default user in *Production*

PR #69, opened 2015-04-13
Rebuild migrations to fix broken migration

wasabideveloper:
Django was missing columns after squash...


njordan-hsr:
The old (deleted though this pull request) migrations run cleanly on my machine, but IMO there isn't a reason to keep this migration, since it seems to cause issues - we still have history preserved through git, if we ever wish to do this thing on "production" data (which we hopefully do not need!).

wasabideveloper:
Do not merge this. There are problems with null constraints.

PR #82, opened 2015-04-14
avoid bash backticks

das-g:


developmentEnvironment/README.md
```diff
@@ -108,12 +108,12 @@ You need to specify the ip. Otherwise you are not able to reach the application
 ```shell
 #!/bin/bash
 
-CURRENTDIR=`dirname $0`
```
wasabideveloper: Why is it needed, to avoid backticks?

das-g: For the current code, it's not needed, good old backticks do exactly the same for it. But this changes makes the code friendlier to modification, as the more modern `$()` handles whitespace and special characters a bit saner and can be nested naturally without headaches about escaping the inner backticks.

PR #85, opened 2015-04-17
Extraction order status page & extraction order creation confirmation page

wasabideveloper:


source/excerptexport/templates/excerptexport/templates/extraction_order_status.html
```diff
@@ -0,0 +1,54 @@
...
+                    </td>
+                </tr>
+                {% if extraction_order.process_start_date %}
+                <tr><th>Extraction started at</th><td>{{ extraction_order.process_start_date|date:"r" }}</td>
+                {% endif %}
+                <tr><th>Issuer</th><td>{{ extraction_order.orderer.get_username }}</td></tr>
+        </body>
+    </table>
+
+    {% if extraction_order.get_state_display == 'FINISHED' %}
```
das-g: Is it OK to have this level of business logic in django templates? (The knowledge that there are only downloads for finished orders isn't a pure view matter.) @njordan-hsr, how should one handle stuff like this?

source/excerptexport/views.py
```diff
@@ -80,18 +81,28 @@ def create_excerpt_export(request):
...
+
     view_context['use_existing'] = 'existing_excerpt_id' in vars()  # TODO: The view should not have to know
     export_options = get_export_options(request.POST, settings.EXPORT_OPTIONS)
     view_context['options'] = export_options
 
     trigger_data_conversion(extraction_order, export_options)
 
-    return render(request, 'excerptexport/templates/create_excerpt_export.html', view_context)
+    #response = HttpResponse(loader.get_template('excerptexport/templates/create_excerpt_export.html').render(RequestContext(request, view_context)))
+    #    render(request, 'excerptexport/templates/create_excerpt_export.html', view_context)
```
das-g: Can the commented-out code be removed?

source/excerptexport/views.py
```diff
@@ -80,18 +81,28 @@ def create_excerpt_export(request):
...
+    response = render_to_response('excerptexport/templates/create_excerpt_export.html', view_context, context_instance=RequestContext(request))
+#response['Cache-Control'] = 'no-cache'
+#return response
```
das-g: Can the commented-out code be removed?

source/excerptexport/templates/excerptexport/templates/extraction_order_status.html
njordan-hsr: I think the display-logic is OK in templates, like the color, but would be nicer if it is extracted into (reusable) snippets (`partials`) and have a clear name.
For easier use, the [with-tag](https://docs.djangoproject.com/en/1.8/ref/templates/builtins/#with) could be useful here, to not have to rely on a specific object being present.

But the business logic should really be extracted, into the view or the model, if the model keeps the information. The template (the `view` in MVC) should really only be about displaying information.

Issue #14, opened 2015-03-10
Integrate OpenID authentication for Django


wasabideveloper:

wasabideveloper:
Estimation: ~1.5d

PR #99, opened 2015-05-04
Feature/integrate open id authentication #14

njordan-hsr:
* Login with google oauth
* Some cleanup
* Circular dependencies resolved with the urls.py

**Warning**: Login in Settings with oauth for testing purposes only and localhost only!

Tested Application:
```sh
./manage.py check
./manage.py test
```
Application manually Tested:
* login through google oauth
* get permission denied for views
* add rights as admin to user
* access views with permissions

@das-g & @wasabideveloper please review

### DIFF
--- a/source/excerptexport/urls.py
+++ b/source/excerptexport/urls.py
@@ -1,21 +1,25 @@
 from django.conf.urls import patterns, url
+from excerptexport.views import index, access_denied, new_excerpt_export, show_downloads, download_file, \
+    create_excerpt_export, extraction_order_status
 
 
-urlpatterns = patterns(
-    'excerptexport.views',
-    url(r'^$', 'index', name='index'),
-    url(r'^access_denied/$', 'access_denied', name='access_denied'),
-    url(r'^new/$', 'new_excerpt_export', name='new'),
-    url(r'^downloads/$', 'show_downloads', name='downloads'),
-    url(r'^download/$', 'download_file', name='download'),
-    url(r'^create/$', 'create_excerpt_export', name='create'),
-    url(r'^orders/(?P<extraction_order_id>[0-9]+)$', 'extraction_order_status', name='status')
+except_export_urlpatterns = patterns(
+    '',
+    url(r'^$', index, name='index'),
+    url(r'^access_denied/$', access_denied, name='access_denied'),
+    url(r'^new/$', new_excerpt_export, name='new'),
+    url(r'^downloads/$', show_downloads, name='downloads'),
+    url(r'^download/$', download_file, name='download'),
+    url(r'^create/$', create_excerpt_export, name='create'),
+    url(r'^orders/(?P<extraction_order_id>[0-9]+)$', extraction_order_status, name='status')
 )
 
-urlpatterns += patterns(
+login_logout_patterns = patterns(
     '',
     url(r'^login/$', 'django.contrib.auth.views.login',
         {'template_name': 'excerptexport/templates/login.html'}, name='login'),
     url(r'^logout/$', 'django.contrib.auth.views.logout',
         {'template_name': 'excerptexport/templates/logout.html'}, name='logout'),
 )
+
+urlpatterns = except_export_urlpatterns + login_logout_patterns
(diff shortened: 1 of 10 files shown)


source/excerptexport/urls.py
```diff
@@ -1,23 +1,21 @@
 from django.conf.urls import patterns, url
 
-from excerptexport import views
-
 
 urlpatterns = patterns(
-    '',
-    url(r'^$', views.index, name='index'),
```
wasabideveloper: I'm not a fan of 'magic' string references instead of real references. What's the reason using string references? Circular references? Can't python handle this?

source/excerptexport/views.py
```diff
@@ -20,8 +21,27 @@
...
+    excerpt_export_permissions = [
+        'excerptexport.add_boundinggeometry',
+        'excerptexport.change_boundinggeometry',
+        'excerptexport.add_excerpt',
+        'excerptexport.change_excerpt',
+        'excerptexport.add_extractionorder',
+        'excerptexport.change_extractionorder',
+    ]
+    # This is a simple hack, instead of displaying a login page, it shows a permission denied page if the user is
+    # logged in
```
wasabideveloper: The PD-page is shown, if the user is logged in or not? (comment correct?).

source/osmaxx/settings.py
```diff
@@ -112,6 +190,31 @@
...
+SOCIAL_AUTH_NEW_USER_REDIRECT_URL = '/excerptexport/'
+# Like SOCIAL_AUTH_NEW_USER_REDIRECT_URL but for new associated accounts (user is already logged in).
+# Used in place of SOCIAL_AUTH_LOGIN_REDIRECT_URL
+SOCIAL_AUTH_NEW_ASSOCIATION_REDIRECT_URL = '/excerptexport/'
+# The user will be redirected to this URL when a social account is disconnected
+SOCIAL_AUTH_DISCONNECT_REDIRECT_URL = '/excerptexport/'
+# Inactive users can be redirected to this URL when trying to authenticate.
+# Successful URLs will default to SOCIAL_AUTH_LOGIN_URL while error URLs will fallback to SOCIAL_AUTH_LOGIN_ERROR_URL.
+SOCIAL_AUTH_INACTIVE_USER_URL = '/excerptexport/'
+SOCIAL_AUTH_LOGIN_SUCCESS_URL = '/excerptexport/'
```
wasabideveloper: Is it possible to move the path to a variable instead of repeating it again and again? We will refactor the url structure soon and a variable will simplify that.

source/osmaxx/urls.py
```diff
@@ -10,4 +10,6 @@
 
     url(r'^excerptexport/', include('excerptexport.urls', namespace='excerptexport')),
     url(r'^admin/', include(admin.site.urls)),
+
+    url('', include('social.apps.django_app.urls', namespace='social')),
```
wasabideveloper: We should use string references or real references but not mix it. (admin/)?

source/excerptexport/urls.py
njordan-hsr: Circular references indeed.

source/excerptexport/views.py
njordan-hsr: Comment is correct.

source/osmaxx/urls.py
njordan-hsr: agreed, but that's not a line from me, you're commenting your own doing ;-)

source/osmaxx/settings.py
njordan-hsr: They should not be the same, and therefore I'd advise against a variable.

source/osmaxx/urls.py
wasabideveloper: I commented your line, because you had a 50% decision to use strings or variables ;-) and you refactored all urls in the other file. so we should do it here too...

njordan-hsr: :-) Done

source/excerptexport/urls.py
njordan-hsr: reverse_lazy is your friend though, don't need the string references anymore, hence -> :-)

njordan-hsr: reverse_lazy is your friend though, don't need the string references anymore, hence -> :-)

wasabideveloper: Great!

Issue #28, opened 2015-03-10
Show user all available downloads ordered by extraction order


wasabideveloper:

das-g:
- [ ] Show pending requests

wasabideveloper:
* Order by extraction order

wasabideveloper:
Estimation: ~0.5d

PR #101, opened 2015-05-04
Feature/order downloads by extraction order#28

wasabideveloper:
- List downloads ordered by extraction order

Application tested:
- ran check
- ran tests
- Checked by hand:
  - /downloads
  - /downloads/{uuid}


njordan-hsr:
Closes #28.

PR #106, opened 2015-05-04
Feature/list extraction orders#103

wasabideveloper:
Application tested:
- ran check
- ran test
- tested by hand:
  - /orders
  - /orders/{id}
  - /downloads
  - /downloads/{uuid}

### DIFF
--- a/source/excerptexport/templates/excerptexport/templates/extraction_order_status.html
+++ b/source/excerptexport/templates/excerptexport/templates/extraction_order_status.html
@@ -29,7 +29,7 @@
 
         <a href="mailto:?subject=Download%20map%20data%20of%20{{ extraction_order.excerpt.name }}&body=
 {% for file in extraction_order.output_files.all %}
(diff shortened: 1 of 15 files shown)


source/excerptexport/templates/excerptexport/templates/extraction_order_status.html
```diff
@@ -4,7 +4,8 @@
 
 {% block main %}
     <nav>
-        <a href="{% url 'excerptexport:index' %}">&#8810; Home</a>
```
njordan-hsr: it might be a goo idea to consider making a breadcrumb instead of writing this everywhere. I'll file an issue, isn't urgent.

source/excerptexport/views.py
```diff
@@ -178,7 +178,7 @@ def get_export_options(requestPostValues, optionConfig):
     return export_options
 
 
-@login_required(login_url='/excerptexport/login/')
+@login_required()
```
njordan-hsr: it would be enough to wirte `@login_required`, no parentheses - but I like explicit better, so unless this is bothering in any way, I'd leave it as is.

Issue #51, opened 2015-03-23
Support administrative area extraction orders


das-g:
#21 (pull request #50) only handles extraction orders with
- newly created excerpts
- existing user-defined excerpts

Choosing a predefined country/region (administrative area) from the list and clicking 'Create excerpt export request') results in the following error:
```
Environment:


Request Method: POST
Request URL: http://127.0.0.1:8000/excerptexport/create/

Django Version: 1.7.4
Python Version: 3.4.0
Installed Applications:
('django.contrib.admin',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'django.contrib.gis',
 'excerptexport')
Installed Middleware:
('django.contrib.sessions.middleware.SessionMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'django.contrib.auth.middleware.SessionAuthenticationMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware')


Traceback:
File "/var/www/eda/environment/lib/python3.4/site-packages/django/core/handlers/base.py" in get_response
  111.                     response = wrapped_callback(request, *callback_args, **callback_kwargs)
File "/var/www/eda/environment/lib/python3.4/site-packages/django/contrib/auth/decorators.py" in _wrapped_view
  21.                 return view_func(request, *args, **kwargs)
File "/var/www/eda/projects/excerptexport/views.py" in create_excerpt_export
  45.             orderer = request.user
File "/var/www/eda/environment/lib/python3.4/site-packages/django/db/models/manager.py" in manager_method
  92.                 return getattr(self.get_queryset(), name)(*args, **kwargs)
File "/var/www/eda/environment/lib/python3.4/site-packages/django/db/models/query.py" in create
  372.         obj.save(force_insert=True, using=self.db)
File "/var/www/eda/environment/lib/python3.4/site-packages/django/db/models/base.py" in save
  589.                        force_update=force_update, update_fields=update_fields)
File "/var/www/eda/environment/lib/python3.4/site-packages/django/db/models/base.py" in save_base
  617.             updated = self._save_table(raw, cls, force_insert, force_update, using, update_fields)
[...]

PR #115, opened 2015-05-19
Feature/predefined areas #51

das-g:


das-g:
@njordan-hsr I've merged this without waiting for a review for the release of [milestone Alpha 3](https://github.com/geometalab/osmaxx/milestones/Alpha%203%20-%20Refactoring,%20OAuth2%20integration%20&%20Administrative%20areas%20support). A post-merge review would still be appreciated, off course.

source/excerptexport/models/bounding_geometry.py
```diff
@@ -1,40 +1,67 @@
...
+            self.north_east,
+            GEOSGeometry('POINT(%s %s)' % (self.east, self.south)),
+            self.south_west
         ])
-        bounding_geometry.geometry = GEOSGeometry(polygon)
-        return bounding_geometry
+        return GEOSGeometry(polygon)
 
-    type = enum.EnumField(BoundingGeometryType, default=BoundingGeometryType.BOUNDINGBOX, verbose_name=_('type'))
-    geometry = models.GeometryField(srid=4326, blank=True, null=True, spatial_index=True, verbose_name=_('geometry'))
```
das-g: I had to drop this translation as the `geometry` field is replaced by a method of the same name. If we need it, we might have to look into Django's notion of virtual (i.e. computed) fields, which seems to be based on python properties.

PR #118, opened 2015-06-01
Feature/split deployment container using compose#75

njordan-hsr:
Docker compose enabled.

ran:
* tests (OK)


README.md
```diff
@@ -5,9 +5,46 @@ Cuts out OpenStreetMap data, processes it to geodata and converts it to typical
...
+From now on you can start the project using docker compose:
+
+`docker-compose up webapp`
+
+### Running things inside the container
+`docker-compose run webapp <command>`
+
+For example, to run tests, you can use:
+
+`docker-compose run webapp python3 manage.py test`
```
das-g: I assume we don't use `virtualenv` in the containers, because each container will host at most one python application, anyway, and is thus itself isolation enough.

database/initdb-postgis.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+POSTGRES="gosu postgres postgres"
```
das-g: WTF is `gosu`? Is that an alternative to `su`/`sudo`?

webapp/Dockerfile
```diff
@@ -0,0 +1,39 @@
+FROM debian:jessie
+ENV REFRESHED_AT 2015-05-18
+
+# update packages and prepare to build software
+RUN ["apt-get", "update"]
+RUN ["apt-get", "-y", "install", "build-essential"]
+
+ENV LANG en_US.utf8
+
+RUN ["apt-get", "-y", "install", "git"]
```
das-g: Why install `git` in the container? Will it be able to access the repo?

webapp/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+
+# install latest python
+RUN ["apt-get", "-y", "install", "python3-dev", "python3-pip", "python3-pil", "python3-psycopg2"]
+
+# prepare postgreSQL support
+RUN ["apt-get", "-y", "install", "postgresql-server-dev-all"]
+RUN ["apt-get", "-y", "install", "binutils", "libproj-dev", "gdal-bin", "libgeoip1", "gdal-bin", "python-gdal"]
```
das-g: `gdal-bin` probably shouldn't be twice in here.

webapp/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+
+ENV LANG en_US.utf8
+
+RUN ["apt-get", "-y", "install", "git"]
+
+# install latest python
+RUN ["apt-get", "-y", "install", "python3-dev", "python3-pip", "python3-pil", "python3-psycopg2"]
+
+# prepare postgreSQL support
+RUN ["apt-get", "-y", "install", "postgresql-server-dev-all"]
```
das-g: Shouldn't the webapp container only need an SQL client?

webapp/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+WORKDIR /home/osmaxx
+RUN ["chown", "-R", "osmaxx:osmaxx", "/home/osmaxx"]
+# ADD must be after chown see http://stackoverflow.com/a/26145444/1281947
+ADD ./osmaxx/requirements /tmp/requirements/
+# manage requirements
+ENV REQUIREMENTS_REFRESHED_AT 2015-05-16
+# change to "requirements/production.txt" on production
+RUN ["pip3", "install", "-r", "/tmp/requirements/local.txt"]
+ENV DATABASE_URL postgis://postgres@database/postgres
+# uncomment the line below to use container as a non-root user
```
das-g: What line does this comment apply to?

webapp/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+RUN ["apt-get", "-y", "install", "git"]
+
+# install latest python
+RUN ["apt-get", "-y", "install", "python3-dev", "python3-pip", "python3-pil", "python3-psycopg2"]
+
+# prepare postgreSQL support
+RUN ["apt-get", "-y", "install", "postgresql-server-dev-all"]
+RUN ["apt-get", "-y", "install", "binutils", "libproj-dev", "gdal-bin", "libgeoip1", "gdal-bin", "python-gdal"]
+
+# move into our working directory
```
das-g: What line does this comment describe?

webapp/Dockerfile
njordan-hsr: Needed for installing through pip, when using a repository. Should probably be removed, since we don't use any repositories right now.

database/initdb-postgis.sh
njordan-hsr: https://docs.docker.com/articles/dockerfile_best-practices/ -> search for gosu

README.md
njordan-hsr: Yes, that is also the default behavior of the 'official' python images from docker.

webapp/Dockerfile
njordan-hsr: will be fixed soon.

webapp/Dockerfile
njordan-hsr: yes, in an ideal world, it would, but psycopg2 is being compiled, which needs the headers. Will be fixed in the new setup as well.

webapp/Dockerfile
njordan-hsr: will be removed :+1:

webapp/Dockerfile
njordan-hsr: will be removed :+1:

webapp/osmaxx/config/settings/common.py
```diff
@@ -0,0 +1,329 @@
...
+
+
+# MEDIA CONFIGURATION
+# ------------------------------------------------------------------------------
+# See: https://docs.djangoproject.com/en/dev/ref/settings/#media-root
+
+# data & media
+
+MEDIA_ROOT = str(APPS_DIR('media'))
+PRIVATE_MEDIA_ROOT = str(APPS_DIR.path('media', '..', 'private-media'))
```
das-g: As you've changed these constants, please re-generate migrations that depend on them.

osmaxx/config/settings/common.py
```diff
@@ -0,0 +1,329 @@
...
+
+
+# MEDIA CONFIGURATION
+# ------------------------------------------------------------------------------
+# See: https://docs.djangoproject.com/en/dev/ref/settings/#media-root
+
+# data & media
+
+MEDIA_ROOT = str(APPS_DIR('..', 'media'))
+PRIVATE_MEDIA_ROOT = str(APPS_DIR.path('..', 'private_media'))
```
das-g: As you've changed these constants, please re-generate migrations that depend on them.

njordan-hsr: good catch, thanks

Issue #93, opened 2015-04-21
Implement Celery invoking data conversion process


wasabideveloper:
#34, #33

### Steps

* [x] create an celery worker process (with docker images)
* [x] create a simple start and response worker task

PR #120, opened 2015-06-08
Feature/implement celery invoking data conversion process #93

njordan-hsr:
First step for #93


njordan-hsr:
@das-g Sorry, is a very big commit with many moved files.

PR #126, opened 2015-06-29
Feature/avoid hardcoded file system paths in migrations #54

### LINKED ISSUE #54: Avoid hardcoded file system paths in migrations
Replace path string concatenation with os join path within a function for `upload_to`, so they may differ between deployments.

njordan-hsr:
Closes #54


njordan-hsr:
Closes #54

PR #143, opened 2015-07-20
Feature/#12 provide initial access groups and permissions

njordan-hsr:
resolves #12.

* tested manually:
    * access to all protected pages with:
        * a valid user
        * an invalid user (no access, but login)
        * admin
        * anonymous (not logged in)
* ran test

@das-g @wasabideveloper please review.

PR #149, opened 2015-07-22
user permissions instead of querying groups

das-g:


osmaxx-py/osmaxx/contrib/auth/migrations/0002_add_default_usergroup_osmaxx.py
```diff
@@ -8,7 +8,13 @@
...
     """Add group osmaxx."""
     Group = apps.get_model("auth", "Group")
-    Group.objects.create(name=settings.OSMAXX_FRONTEND_USER_GROUP)
+    Permission = apps.get_model("auth", "Permission")
+    ContentType = apps.get_model("contenttypes", "ContentType")
+    ExtractionOrder = apps.get_model("excerptexport", "ExtractionOrder")
+    group = Group.objects.create(name=settings.OSMAXX_FRONTEND_USER_GROUP)
+    content_type = ContentType.objects.get_for_model(ExtractionOrder)
+    permission, created = Permission.objects.get_or_create(codename='add_extractionorder', content_type=content_type)
+    group.permissions.add(permission)
```
das-g: Now that rest of `feature/#12_provide-initial-access-groups-and-permissions` got already merged in #143, should the adding of the relationship from the group to the permission be in a separate migration?

osmaxx-py/osmaxx/contrib/auth/migrations/0002_add_default_usergroup_osmaxx.py
```diff
@@ -8,7 +8,13 @@
 def update_site_forward(apps, schema_editor):
     """Add group osmaxx."""
     Group = apps.get_model("auth", "Group")
-    Group.objects.create(name=settings.OSMAXX_FRONTEND_USER_GROUP)
+    Permission = apps.get_model("auth", "Permission")
+    ContentType = apps.get_model("contenttypes", "ContentType")
+    ExtractionOrder = apps.get_model("excerptexport", "ExtractionOrder")
+    group = Group.objects.create(name=settings.OSMAXX_FRONTEND_USER_GROUP)
+    content_type = ContentType.objects.get_for_model(ExtractionOrder)
```
das-g: As we get **or create** a permission, should we remove it again in the backwards migration? (Probably not, as it'd be recreated by the `post_migrate` signal handler, anyway.)

osmaxx-py/osmaxx/contrib/auth/migrations/0002_add_default_usergroup_osmaxx.py
njordan-hsr: Would be an argument for it, but I think this is just fine: backwards migrations would be much harder when having separate migrations.

osmaxx-py/osmaxx/contrib/auth/migrations/0002_add_default_usergroup_osmaxx.py
njordan-hsr: That is correct, AFAIK we only need to remove the group.

PR #156, opened 2015-07-29
Feature/excerpt converters#123

### LINKED ISSUE #123: Have dummy export converter (pipeline) declare supported export formats
* [ ] Move dummy pipleline to Django app.
* [ ] Have that app register its supported export formats. (See decision #15)
* [ ] Make the web interface display registered export formats instead of ones semi-soft-coded in settings.

Expected duration: ~5d.

wasabideveloper:
Implementation of feature #123 (Merge will close #123 ).

* locales **NOT** compiled -> do this at the end of the project some time (prevent huge locale diffs in changes)
* ran flake8
* ran check
* ran test
* tested views by hand:
  * /orders/new [get/post]
  * /orders/{id}

To be reviewed by:
- [x] @das-g 
- [ ] @njordan-hsr


osmaxx-py/config/settings/common.py
```diff
@@ -50,7 +50,12 @@
 LOCAL_APPS = (
     'osmaxx.excerptexport',
     'osmaxx.social_auth',
+    'excerptconverter.dummyexcerptconverter'
```
njordan-hsr: wouldn't 
`excerptconverter.dummy`
be enough namespacing?

osmaxx-py/excerptconverter/tests/__init__.py
```diff
@@ -0,0 +1 @@
+# scripts with names of form 'test*.py' are autodiscovered
```
njordan-hsr: Why is this comment here? I don't think we should document django documentation/python test framework inside here.

wasabideveloper: It's your code or the code of @das-g I copied from the other test module and I think it makes sense, because it's not the normal way python works. "tests" needs to be a python module but it's not necessary to import the tests = it's not the way python works normal.

osmaxx-py/config/settings/common.py
wasabideveloper: You mean: Rename excerptconverter/dummyexcerptconverter to excerptconverter/dummy?

osmaxx-py/excerptconverter/tests/__init__.py
das-g: I'd leave that comment in there, for the reasons @wasabideveloper mentioned.

osmaxx-py/excerptconverter/baseexcerptconverter/base_excerpt_converter.py
```diff
@@ -0,0 +1,89 @@
...
+from django.core.mail import send_mail
+from django.utils.translation import ugettext_lazy as _
+
+
+class BaseExcerptConverter(metaclass=abstract_base_class.ABCMeta):
+    available_converters = []
+    name = 'base'
+    export_formats = {}
+    export_options = {}
+    steps_total = 0
```
das-g: Only `available_converters` is needed in the abstract base class. The other class fields here should be defined in concrete derived classes, only, _not_ in the abstract base class.

To document what the base class requires of its concrete derivatives, a [Python Docstring](https://www.python.org/dev/peps/pep-0257/) is the way to go, rather than dead code.

osmaxx-py/excerptconverter/baseexcerptconverter/base_excerpt_converter.py
```diff
@@ -0,0 +1,89 @@
...
+        if email:
+            if hasattr(user, 'email'):
+                send_mail(
+                    '[OSMAXX] '+message_text,
+                    message_text,
+                    'no-reply@osmaxx.hsr.ch',
+                    [user.email]
+                )
+            else:
+                stored_messages.api.add_message_for(
```
das-g: Why not call `inform_user(user, messages.WARNING, _("There is no email address assigned to your account. " + "You won't be notified by email on process finish!", email=False)` here?

osmaxx-py/excerptconverter/baseexcerptconverter/base_excerpt_converter.py
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+    @staticmethod
+    @abstract_base_class.abstractmethod
+    @shared_task
+    def execute_task(extraction_order_id, supported_export_formats, execution_configuration):
+        return None
+
+    @staticmethod
+    def inform_user(user, message_type, message_text, email=True):
```
das-g: Would it make sense to move this function to the user class and make it an instance method?

osmaxx-py/excerptconverter/baseexcerptconverter/base_excerpt_converter.py
```diff
@@ -0,0 +1,89 @@
...
+                },
+                'markdown': {
+                    'name': 'Markdown',
+                    'file_extension': 'md',
+                    'mime_type': 'text/markdown'
+                }
+            }
+        """
+        # queue celery task
+        # run_as_celery_tasks=False allows to run as normal functions for testing
```
das-g: These facts are relevant to users of the method, thus should probably be part of the Docstring.

osmaxx-py/excerptconverter/converter_manager.py
```diff
@@ -0,0 +1,39 @@
+from excerptconverter.baseexcerptconverter import BaseExcerptConverter
+
+
+class ConverterManager:
+    @staticmethod
+    def converter_configuration():
+        export_options = {}
+        for Converter in BaseExcerptConverter.available_converters:
+            export_options[Converter.__name__] = Converter.converter_configuration()
+        return export_options
```
das-g: This function can be rewritten as a single [dictionary comprehension](https://www.python.org/dev/peps/pep-0274/):

    @staticmethod
    def converter_configuration():
        return {Converter.__name__: Converter.converter_configuration()
                for Converter in BaseExcerptConverter.available_converters}

osmaxx-py/osmaxx/excerptexport/forms/export_options_form.py
```diff
@@ -1,16 +1,18 @@
...
-    def __init__(self, *args, **kwargs):
+    def __init__(self, converter_configuration, *args, **kwargs):
         super(ExportOptionsForm, self).__init__(*args, **kwargs)
-        for export_option_key, export_option in settings.EXPORT_OPTIONS.items():
+        self.converter_configuration = converter_configuration
+
+        for export_option_key, export_option in self.converter_configuration.items():
             self.create_checkboxes(
-                'export_options.'+export_option_key+'.formats', export_option['name'],
+                'export_options.'+export_option_key+'.formats',
```
das-g: Put spaces around operator `+`. (Doesn't `flake8` complain about the missing spaces?)

das-g: Or even better, use string interpolation, like in the line below.

osmaxx-py/osmaxx/excerptexport/tests/test_models.py
```diff
@@ -40,3 +46,37 @@ def test_geometry_has_clockwise_exterior_ring_starting_at_south_west(self):
         BBoxBoundingGeometry.create_from_bounding_box_coordinates(1.1, 2.2, 3.3, 4.4)
         self.assertEqual(BoundingGeometry.objects.first().bboxboundinggeometry.geometry.exterior_ring[:],
                          [(4.4, 3.3), (4.4, 1.1), (2.2, 1.1), (2.2, 3.3), (4.4, 3.3)])
+
+
+class ExtractionOrderTestCase(TestCase):
+    user = None
+    excerpt = None
+    extraction_configuration = None
```
das-g: No, no, no. This isn't C or Java.

osmaxx-py/osmaxx/excerptexport/tests/test_models.py
```diff
@@ -40,3 +46,37 @@ def test_geometry_has_clockwise_exterior_ring_starting_at_south_west(self):
...
+        }
+
+    def test_create_and_restore_extraction_configuration(self):
+        extraction_order_id = models.ExtractionOrder.objects.create(
+            excerpt=self.excerpt,
+            orderer=self.user,
+            extraction_configuration={'formats': ['txt'], 'options': {'detail_level': 'standard'}}
+        ).id
+        extraction_order = models.ExtractionOrder.objects.get(pk=extraction_order_id)
+        self.assertEqual(extraction_order._extraction_configuration, json.dumps(self.extraction_configuration))
```
das-g: Aren't you testing implementation details instead of contracts with that first assertion? I guess `ExtractionOrder._extraction_configuration` isn't meant to be used from outside the class, so the second assertion should suffice to ensure the class' contract.

PR #157, opened 2015-07-30
Feature/shared data only container

njordan-hsr:
A simple data only container for shared data.

Ran tests, OK.


wasabideveloper:
I Added a test script calling the application tests and testing the volume mount.
Please review again:
- [ ] @das-g 
- [ ] @njordan-hsr

Issue #163, opened 2015-08-05
Move database data to data-only container


wasabideveloper:
- [x] Create /database volume on the shared-data container.
- [x] Mount this volume as database data volume into the database container.
- [x] Disable network for data only container

**Estimated duration**: ~1d

das-g:
closed with #171

das-g:
closed with #171

PR #171, opened 2015-08-10
Feature/#163 move database data to data only container

njordan-hsr:
Implements #163 .

Ran tests: OK.
Ran checks: OK.

Reviewed by:
* [x] @das-g 
* [x] @wasabideveloper


test.sh
```diff
@@ -7,52 +7,131 @@ GREEN="\e[32m"
 MAGENTA="\e[95m"
 RESET="\e[0m"
 
+# This function will be called when running this script
+function main(){
+    # This function will be called when running this script
```
wasabideveloper: Redundant comment (see 2 lines before)

test.sh
```diff
@@ -7,52 +7,131 @@ GREEN="\e[32m"
...
 
+    dcompose run $CELERY_CONTAINER /bin/bash -c "rm $TEST_FILE" &> test.log
+    if [ $? -ne 0 ]; then
+        echo -e "${RED}Test file clean up failed ${RESET}"
+    fi
+}
 
-# docker volume configuration tests
+function persisting_database_data_tests() {
+    # clean containers needed for testing this...
```
wasabideveloper: Please implement a set_up and tear_down instead. Test should not need to know if the container was used before...

test.sh
```diff
@@ -7,52 +7,131 @@ GREEN="\e[32m"
...
 
+# This function will be called when running this script
+function main(){
+    # This function will be called when running this script
+    run_tests_in_development;
+    run_tests_in_production;
+}
 
-# development or production?
-if [[ $(ls -l docker-compose.yml) == *"production.yml"* ]]; then
```
wasabideveloper: I think, you miss understood the function of this. This should run production tests if you are in production mode and development tests, if the application is in development mode. Maybe it's useful, to run all tests in development mode to test production configuration too. But development tests should not be run on productive systems.

test.sh
```diff
@@ -7,52 +7,131 @@ GREEN="\e[32m"
...
+    dcompose rm -f ${CONTAINER_TO_BE_RESETTED} &> test.log;
+    dcompose build ${CONTAINER_TO_BE_RESETTED} &> test.log;
+}
 
-# application tests
-echo -e "${MAGENTA}-------------------"
-echo -e "Application checks:"
-echo -e "-------------------${RESET}"
-docker-compose run $WEBAPP_CONTAINER /bin/bash -c "python3 manage.py check"
+function dcompose() {
```
wasabideveloper: Please use full speaking names and no shortcuts. Better: `run_docker_compose`

test.sh
```diff
@@ -7,52 +7,131 @@ GREEN="\e[32m"
...
-if [[ $(ls -l docker-compose.yml) == *"production.yml"* ]]; then
+function run_tests_in_development() {
+    echo -e "${MAGENTA}"
+    echo -e "=== Development mode ==="
+    echo -e "${RESET}"
+    WEBAPP_CONTAINER="osmaxxwebappdev"
+    CELERY_CONTAINER="osmaxxcelerydev"
+    DB_CONTAINER="osmaxxdatabasedev"
+    COMPOSE_FILE="compose-development.yml"
+    run_tests;
```
wasabideveloper: Please use empty lines to make the code better readable.

test.sh
```diff
@@ -7,52 +7,131 @@ GREEN="\e[32m"
...
+
+function reset_containers() {
+    echo -e "resetting all containers";
+    dcompose stop -t 0 &> test.log;
+    dcompose rm -f &> test.log;
+    dcompose build &> test.log;
+}
 
+function reset_container() {
+    echo -e "resetting container ${CONTAINER_TO_BE_RESETTED};"
```
wasabideveloper: Please extract global variable to function param

test.sh
njordan-hsr: good idea.

test.sh
njordan-hsr: good idea.

test.sh
njordan-hsr: It is not running anything, so I guess `docker_compose` is a better choice?

wasabideveloper: :+1:

test.sh
njordan-hsr: so you prefer this:
```
function reset_container() {
    echo -e "resetting container ${1};
```
over the above?

test.sh
njordan-hsr: I did understand that line, but not why exactly it is needed. It is simple to re-enable, and since it seems to ease deployment, this seems reasonable.

test.sh
wasabideveloper: Am I right, ${CONTAINER_TO_BE_RESETTED} is a global variable and not a named param? In this case, I would prefer
```shell
function reset_container() {
    echo -e "resetting container ${1};
```
Called by 
```shell
reset_container "some_container_name"
```

test.sh
wasabideveloper: If we define, one should not run tests on productive systems, we should use this line to prevent test running on productive systems and print a message.

test.sh
das-g: `docker_compose` can easily be confused with `docker-composie`, which the former is wrapping.

test.sh
das-g: On [line 128](https://github.com/geometalab/osmaxx/pull/171/files#diff-1b531c5e405e0bc95890d52cf2f2f5b5R128), it's used as an environment variable for only that command, so I guess one could call it a 'named parameter'.

wasabideveloper: :-1: I prefer the explicit way using ${1}

njordan-hsr: It is an environment variable, which is set, but can not be accessed outside the script, since it is not exported. But if by global you mean 'available outside the scope of the `reset_container` function', yes. But it is not, in a programming sense, a global variable, but it is not a parameter either. TBH, I prefer this style, but don't care enough to blow it up to a _principles_ discussion :wink: .

PR #201, opened 2015-09-18
fix: containers can't be found when using run and docker-compose

### LINKED ISSUE #193: Fix failing removal if temporary containers of excerpt container chain
This happens in production mode:
containers2
like in development mode:
containers

This happens on the live server too.

Estimated duration: ~2d.

njordan-hsr:
Closes #193 

Ran tests, all green except migrations retained, but this is the same result as on develop and should be fixed separately.

Reviewed by:

* [x] @das-g 
* [x] @wasabideveloper

### DIFF
--- a/osmaxx-py/excerptconverter/gisexcerptconverter/gis_excerpt_converter.py
+++ b/osmaxx-py/excerptconverter/gisexcerptconverter/gis_excerpt_converter.py
@@ -91,7 +91,7 @@ class GisExcerptConverter(BaseExcerptConverter):
             index = 0
             if export_format_key in execution_configuration['formats']:
                 index += 1
-                extraction_command = "docker-compose run excerpt python excerpt.py %(bbox_args)s -f %(format)s" % {
+                extraction_command = "docker-compose run --rm excerpt python excerpt.py %(bbox_args)s -f %(format)s" % {  # noqa
                     'bbox_args':  bbox_args,
                     'format': export_format_key
                 }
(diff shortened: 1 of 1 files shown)


wasabideveloper:
There should be no migration errors on develop. Please try to pull the newest comits.

njordan-hsr:
I did pull the most recent changes. Note, it is not the migrations themselves failing, but the retaining (the last testing step). I'm working on my private machine, that I had to setup first. Maybe I'm using an older version of docker/docker-compose, which leads to that, will check on Monday.

osmaxx-py/excerptconverter/gisexcerptconverter/gis_excerpt_converter.py
```diff
@@ -91,7 +91,7 @@ def extract_excerpts(execution_configuration, extraction_order, bbox_args, conve
             index = 0
             if export_format_key in execution_configuration['formats']:
                 index += 1
-                extraction_command = "docker-compose run excerpt python excerpt.py %(bbox_args)s -f %(format)s" % {
+                extraction_command = "docker-compose run --rm excerpt python excerpt.py %(bbox_args)s -f %(format)s" % {  # noqa
```
das-g: What `flake` message is being suppressed here?

njordan-hsr: line too long. I tried with line breaking, but it just gets worse from a readability standpoint.

PR #207, opened 2015-09-25
Feature/190 create e2e tests

### LINKED ISSUE #190: create end-to-end test
Test creation of extraction order - and check if the conversion completed successful.

Estimated duration: ~2d

njordan-hsr:
Closes #190 

* Added e2e tests
* Ran tests: OK

Reviewed by:

* [ ] @wasabideveloper 
* [x] @das-g


osmaxx-py/config/settings/common.py
```diff
@@ -84,9 +84,6 @@
 # See: https://docs.djangoproject.com/en/dev/ref/settings/#debug
 DEBUG = env.bool("DJANGO_DEBUG", False)
 
-# See: https://docs.djangoproject.com/en/dev/ref/settings/#template-debug
-TEMPLATE_DEBUG = DEBUG
-
```
das-g: :+1: for removing deprecated options.

README.md
```diff
@@ -87,6 +87,64 @@ Start the containers using docker compose:
...
+
+There's a script to facilitate testing including some docker integration tests:
+
+```shell
+./test.sh
+```
+
+#### Running end to end tests (e2e)
+
+**Warnings**:
```
das-g: Should be a subsection title. If nesting becomes too deep for that, consider a separate document.

README.md
```diff
@@ -87,6 +87,64 @@ Start the containers using docker compose:
...
+**Warnings**: 
+
+* These run for quite a long time (over five Minutes!). 
+* The containers are being destroyed at the beginning (creating a clean state). Resulting in
+    * Only run one test at a time
+    * don't run on production!
+    * don't run docker-compose in parallel
+    * exiting data is being destroyed 
+
+_Prerequisites_: an activated virtualenv!
```
das-g: Should be a subsection title. If nesting becomes too deep for that, consider a separate document.

e2e/__init__.py
```diff
@@ -0,0 +1,4 @@
+import sys
+import os
+
+sys.path.append(os.path.join(__file__))
```
das-g: Is the `e2e` package's `__init__.py` the right place to set this up? Shouldn't it rather be done in the module(s) (i.e. regular python file(s)) that relies/rely on this? (So that they'll work in isolation, too, without having to import the package that they're part of themselves.)

e2e/e2e_tests.py
```diff
@@ -0,0 +1,151 @@
...
+from time import sleep
+import unittest
+import subprocess
+
+try:
+    import requests
+    from bs4 import BeautifulSoup
+except ImportError:
+    requirements_file = os.path.join(os.path.dirname(__file__), 'requirements.txt')
+    subprocess.check_call("pip install -r {}".format(requirements_file).split(' '))
```
das-g: A test that installs its own dependencies? Shouldn't it rather fail with a hint on how the user _can_ install them, if they choose so?

As it is itself written in python, doesn't `pip` come with a python API, so we don't have to make a system call to invoke it?

README.md
njordan-hsr: Warnings should be a subsection is what you mean, right? We could put the entire documentation on testing into a separate file, and have a link to there, if that is what would make more sense.

e2e/e2e_tests.py
```diff
@@ -0,0 +1,151 @@
...
+    docker_compose.create_superuser_for_test(username=ADMIN_USER_FOR_TESTS, password=ADMIN_PASSWORD_FOR_TESTS)
+    sleep(10)
+
+
+def _stop_containers():
+    """
+    stop the containers and then cleanup
+    """
+    # docker_compose.stop()
+    # docker_compose.clean()
```
das-g: With these lines commented out, this method does nothing. It still gets called from `if __name__ == '__main__':`, so it probably _should_ do _something_, I guess?

e2e/__init__.py
njordan-hsr: Correct, I put it in there to start with and then forgot to move it, thanks for the catch.

e2e/e2e_tests.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+class TestE2E(unittest.TestCase):
+    COOKIES = dict()
+    CSRF_TOKEN = ''
+
+    def _login(self, next=None):
+        login_url = 'http://localhost:8000/login/'
```
das-g: Might make sense to extract `http://localhost:8000`, which is repeated in this class a lot. (Then again, string interpolation doesn't look as smooth in python as it does in Ruby, so ... , maybe not.)

e2e/e2e_tests.py
```diff
@@ -0,0 +1,151 @@
...
+        r = requests.post('http://localhost:8000/login/', data=login_data)
+        self.assertEqual(r.status_code, 403)
+
+    def test_login_success_with_csrf_token(self):
+        login = self._login(next='/orders/')
+        r = login['request']
+        soup = make_soup(r.content)
+        self.assertIsNotNone(soup.find(name='a', attrs={'href': '/logout/?next=/'}))
+        self.assertEqual(r.status_code, 200)
+        self._logout()
```
das-g: If one of the assertions fail, the logout will be skipped. Is that wanted? If not, put it into a `finally` clause or into `tearDown`. (For the latter, we might want to put the tests that require logout in `tearDown` into a separate TestCase class.)

e2e/e2e_tests.py
njordan-hsr: I agree that it this is ugly, but out of practicality I think it makes sense:

You need to install a virtualenv with the dependencies and remember to activate them every time you run those tests. Since we're mostly using docker for development, the chances are quite slim that it is activated already. Furthermore, using the ./test.sh, if you had forgotten to activate it beforehand, you'd need to rerun the entire test-suite again (the long running nature is a mere symptom and shouldn't be used as an argument - but it is a symptom we currently can't get rid of).

Should be possible using:
```
import pip
pip.main(['install', package])
```
Will change it.

e2e/e2e_tests.py
```diff
@@ -0,0 +1,151 @@
...
+
+        link = self._make_link(soup.find(attrs={'class': 'container-row content'}).a.attrs['href'])
+
+        r = client.get(link)
+        soup = make_soup(r.content)
+
+        # first link
+        self._download_and_test_zip_contents(
+            client,
+            soup.find(name='ul', attrs={'class': 'download_files'}).find_all('a')[0].attrs['href']
```
das-g: This looks really fragile. That we already have to iterate the indices with steps of `2` indicates that this test might arbitrarily break for minor changes, even when everything is still working fine.

e2e/e2e_tests.py
```diff
@@ -0,0 +1,151 @@
...
+        self._download_and_test_zip_contents(
+            client,
+            soup.find(name='ul', attrs={'class': 'download_files'}).find_all('a')[2].attrs['href']
+        )
+
+        # third link
+        self._download_and_test_zip_contents(
+            client,
+            soup.find(name='ul', attrs={'class': 'download_files'}).find_all('a')[4].attrs['href']
+        )
```
das-g: Can we DRY this? Dunno if a loop would be appropriate.

e2e/e2e_tests.py
njordan-hsr: Forget what I wrote, it's not creating a virtualenv here anyway, so it doesn't make sense at all. Will update and use a message to the user what to do to get it up and running.

README.md
das-g: > _Warnings_ should be a subsection is what you mean, right?

Yes, and so should 'Prerequisites'.

> We could put the entire documentation on testing into a separate file, and have a link to there

Note, there's already [section 'Run tests' in `docs/project-development-environment.md`](https://github.com/geometalab/osmaxx/blob/e7f2dac9d737062dba5c668ffa0936a39fae72c7/docs/project-development-environment.md#run-tests) and [section 'Run tests' in `docs/useful-commands.md`](https://github.com/geometalab/osmaxx/blob/e7f2dac9d737062dba5c668ffa0936a39fae72c7/docs/useful-commands.md#run-tests). Probably their content and the above and howto coverage should go together in a separate file. Where appropriate in the other files, we can link to that file or its sections.

e2e/e2e_tests.py
njordan-hsr: yes, forgot commenting them _in_ them after commenting them _out_

e2e/e2e_tests.py
njordan-hsr: good catch.

e2e/helpers/docker_compose.py
```diff
@@ -0,0 +1,62 @@
+import os
+import shlex
+import subprocess
+from time import sleep
+
+
+def stop():
+    subprocess.check_call("docker-compose stop -t 0".split(' '))
```
das-g: Does python have something equivalent or comparable to Ruby's [word array literal notation](http://stackoverflow.com/a/1275147/674064) (`%w(every word is an array item of type string)`)? That'd come in handy here.

README.md
njordan-hsr: Good idea. I'll create a new document for testing in general, but maybe not consolidate the testing commands from `useful-commands.py`?

e2e/helpers/docker_compose.py
```diff
@@ -0,0 +1,62 @@
...
+    """
+    superuser_command = """from django.contrib.auth.models import User
+User.objects.create_superuser(username='{username}', password='{password}', email='{email}')
+    """.format(email=email, username=username, password=password)
+    # TODO: this only works on development mode with mounted source volumes
+    filename = os.path.join(os.path.dirname(__file__), '..', '..', 'osmaxx-py', 'create_superuser.py')
+    file = open(filename, mode='w')
+    file.write(superuser_command)
+    file.flush()
+    file.close()
```
das-g: Would a `with` resource clause make sense for this file usage?

e2e/e2e_tests.py
njordan-hsr: Yes, that is so. But I have to go through the HTML-soup, and we don't have any other ways of finding the links. Suggestions welcome.

e2e/e2e_tests.py
njordan-hsr: something like
```python
for link_index in range(0, 5, 2):
    self._download_and_test_zip_contents(
        client,
        soup.find(name='ul', attrs={'class': 'download_files'}).find_all('a')[link_index].attrs['href']
    )
```

e2e/helpers/docker_compose.py
```diff
@@ -0,0 +1,62 @@
...
+    """
+    password is `password` for testing purposes
+    """
+    superuser_command = """from django.contrib.auth.models import User
+User.objects.create_superuser(username='{username}', password='{password}', email='{email}')
+    """.format(email=email, username=username, password=password)
+    # TODO: this only works on development mode with mounted source volumes
+    filename = os.path.join(os.path.dirname(__file__), '..', '..', 'osmaxx-py', 'create_superuser.py')
+    file = open(filename, mode='w')
+    file.write(superuser_command)
```
das-g: Do we have to store this python code in a file to run it within Django (`manage.py`) context? I think I'd prefer piping it into `python manage.py shell`. (Python has quite some capabilities to manage pipes.)

e2e/e2e_tests.py
njordan-hsr: It doesn't matter, if logout isn't called if something fails, since the cookie with the csrftoken will be invalid anyway. We can actually remove that code completely :)

e2e/helpers/html_helpers.py
```diff
@@ -0,0 +1,5 @@
+from bs4 import BeautifulSoup
+
+
+def make_soup(content):
+    return BeautifulSoup(content, 'html.parser')
```
das-g: As we don't encapsulate BeautifulSoup when using it in the test, wouldn't it make sense to have to initialization inline, too, rather than extracted to here? Then it'd at least look familiar to developers who know this library already.

Or is `make_soup` a convention amongst BeatifulSoup users?

e2e/helpers/zip_file_helpers.py
```diff
@@ -0,0 +1,29 @@
...
+                    '.prj',
+                    '.gpkg',
+                    '.csv',
+                    '.txt',
+                    'pdf',
+                    '.odt',
+                    'sqlite',
+                )
+                if zip_info.filename.endswith(endings):
+                    assert_greater(zip_info.file_size, 80)
```
das-g: Would this limit catch files created from empty extracts? (e.g., somewhere in high sea ocean)

e2e/helpers/docker_compose.py
njordan-hsr: unfortunately not. We can shorten it though, only using `.split()` - but that would be less explicit.

e2e/helpers/docker_compose.py
njordan-hsr: Yes, absolutely.

test.sh
```diff
@@ -16,6 +16,32 @@ function main(){
     else
         run_production_tests;
     fi
+
+    # TODO: currently only work on development settings
```
das-g: If you state what's broken/what the current unwanted limitation is, use `FIXME` rather than `TODO`. If you prefer `TODO`, state what should be changed, not what the current state is.

test.sh
```diff
@@ -16,6 +16,32 @@ function main(){
...
+    if [[ $(ls -l docker-compose.yml) == *"development.yml"* ]]; then
+        if [ $RUN_E2E ]; then
+            if [ $RUN_E2E = 'true' ]; then
```
das-g: Do we have to nest the `if`s or can we use a compound condition with lazy evaluation?

test.sh
```diff
@@ -16,6 +16,32 @@ function main(){
...
+                run_e2e_tests;
+            fi
+        fi
+    fi
+}
+
+function create_tmp_virtualenv() {
```
das-g: I think an `_and_activate` is missing in this function name. :-P

e2e/e2e_tests.py
das-g: Then just remove it :smiley:

e2e/helpers/docker_compose.py
das-g: Does `.split()` split on any whitespace (including tabs, newline, NBSP, ...) or only on `' '`?

njordan-hsr: any whitespace.

e2e/helpers/docker_compose.py
njordan-hsr: feel free and try it, I wasn't able to get it to do what I wanted ;)

e2e/helpers/html_helpers.py
njordan-hsr: Don't think that there is a BeatifulSoup'ers community, I think it makes makes sense to inline it.

e2e/helpers/zip_file_helpers.py
njordan-hsr: yes, if there really isn't any data, it would.

test.sh
njordan-hsr: It has never occurred to me to use FIXME, which is muchmore sensible in this context and recognized even by PyCharm and other editors :+1:

test.sh
njordan-hsr: Haven't found a way for lazy evaluation - it is bash, not python, which should probably be changed at some point anyway, but if you know how lazy evaluation in bash works, let me know!

njordan-hsr:
Ran all tests again: green.

njordan-hsr:
@das-g do you think this is ready to be merged?

test.sh
```diff
@@ -16,6 +16,34 @@ function main(){
     else
         run_production_tests;
     fi
+
+    # FIXME: currently only work on development settings
+    if [[ $(ls -l docker-compose.yml) == *"development.yml"* ]]; then
+        if [ $RUN_E2E ]; then
+            if [ $RUN_E2E = 'true' ]; then
```
das-g: Lazy evaluation should be possible with

    if [[ $(ls -l docker-compose.yml) == *"development.yml"* ]] && [ $RUN_E2E ] && [ $RUN_E2E = 'true' ]; then
if I understand http://wiki.bash-hackers.org/commands/classictest#why_you_should_avoid_using_-a_and_-o correctly.

PR #208, opened 2015-09-25
Feature/202 log exceptions and errors

### LINKED ISSUE #202: Log exceptions and errors
https://docs.djangoproject.com/en/1.8/topics/logging/#using-logging

Log exceptions
```python
try:
    # coode in here
except Exception as exception:
    # python will log the stack trace automatically on call of exception()
    # Only use inside try-except
    logging.exception("XYZ is bad and it failed because....")
```
http://www.alexconrad.org/2013/02/loggingexception.html

njordan-hsr:
Closes #202 .
Logging all errors and exceptions in the main flow (not in utilities)

Ran tests: OK

Reviewed by:

* [x] @das-g 
* [x] @wasabideveloper


osmaxx-py/excerptconverter/dummyexcerptconverter/dummy_excerpt_converter.py
```diff
@@ -89,10 +94,11 @@ def execute_task(extraction_order_id, supported_export_formats, execution_config
                 time.sleep(5)
                 wait_time += 5
                 if wait_time > 30:
+                    logger.exception()
```
das-g: Pass a message. (Here, `"Even after waiting for #{wait_time} #{time_unit_abbreviation}, the extraction order #{extraction_order_id} hasn't been created"` is probably more useful than the exception name.)

osmaxx-py/excerptconverter/dummyexcerptconverter/dummy_excerpt_converter.py
```diff
@@ -120,7 +126,7 @@ def execute_task(extraction_order_id, supported_export_formats, execution_config
             # now set the new state (if all files have been processed) and inform the user about the state
             converter_helper.file_conversion_finished()
         except:
-            # TODO: log stack trace
             message_text = _('The Dummy conversion of extraction order "%s" failed.') % extraction_order.id
             converter_helper.inform_user(messages.ERROR, message_text, email=False)
+            logger.exception()
```
das-g: Pass the exception as message.

osmaxx-py/excerptconverter/gisexcerptconverter/gis_excerpt_converter.py
```diff
@@ -115,23 +119,27 @@ def extract_excerpts(execution_configuration, extraction_order, bbox_args, conve
...
                             converter_helper.inform_user(
                                 messages.ERROR,
-                                _('The extraction of "%(file)s" of extraction order "%(order_id)s" failed.') % {
-                                    'file': result_file_name,
-                                    'order_id': extraction_order.id
-                                },
+                                message,
                                 email=False
                             )
+                            logger.error(message)
```
das-g: Log before informing the user.

osmaxx-py/excerptconverter/gisexcerptconverter/gis_excerpt_converter.py
```diff
@@ -115,23 +119,27 @@ def extract_excerpts(execution_configuration, extraction_order, bbox_args, conve
...
                     converter_helper.inform_user(
                         messages.ERROR,
-                        _('The extraction of "%(file_type)s" of extraction order "%(order_id)s" failed.') % {
-                            'file_type': export_format_config['name'],
-                            'order_id': extraction_order.id
-                        },
+                        message,
                         email=False
                     )
+                    logger.error(message)
```
das-g: Log before informing the user.

osmaxx-py/excerptconverter/gisexcerptconverter/gis_excerpt_converter.py
```diff
@@ -223,19 +232,23 @@ def execute_task(extraction_order_id, supported_export_formats, execution_config
...
                 elif type(bounding_geometry) == models.OsmosisPolygonFilterBoundingGeometry:
+                    message = _('GIS excerpt converter is not yet able to extract polygon excerpts.')
                     converter_helper.inform_user(
                         messages.ERROR,
-                        _('GIS excerpt converter is not yet able to extract polygon excerpts.'),
+                        message,
                         email=False
                     )
-
+                    logging.error(message)
```
das-g: Log before informing the user.

osmaxx-py/excerptconverter/gisexcerptconverter/gis_excerpt_converter.py
```diff
@@ -223,19 +232,23 @@ def execute_task(extraction_order_id, supported_export_formats, execution_config
...
+                        type=type(bounding_geometry).__name__
+                    )
                     converter_helper.inform_user(
                         messages.ERROR,
-                        _('GIS excerpt converter is not yet able to extract excerpts of type %s.') %
-                        type(bounding_geometry).__name__,
+                        message,
                         email=False
                     )
+                    logging.error(message)
```
das-g: Log before informing the user.

osmaxx-py/excerptconverter/gisexcerptconverter/gis_excerpt_converter.py
```diff
@@ -173,6 +181,7 @@ def execute_task(extraction_order_id, supported_export_formats, execution_config
                 time.sleep(5)
                 wait_time += 5
                 if wait_time > 30:
+                    logger.exception()
```
das-g: Pass a message. (Here, `"Even after waiting for #{wait_time} #{time_unit_abbreviation}, the extraction order #{extraction_order_id} hasn't been created"` is probably more useful than the exception name.)

osmaxx-py/excerptconverter/gisexcerptconverter/gis_excerpt_converter.py
```diff
@@ -249,11 +262,13 @@ def execute_task(extraction_order_id, supported_export_formats, execution_config
                     },
                     email=False
                 )
+                logger.exception()
```
das-g: Log before informing the user.
Pass the Exception as message.

osmaxx-py/osmaxx/excerptexport/urls.py
```diff
@@ -12,7 +12,7 @@
 )
 
 
-except_export_urlpatterns = [
+excerpt_export_urlpatterns = [
```
das-g: good catch :+1:

osmaxx-py/excerptconverter/gisexcerptconverter/gis_excerpt_converter.py
wasabideveloper: And please log what crazy type of bounding_geometry we got to fall in else (This should never happen :wink: ).

osmaxx-py/excerptconverter/gisexcerptconverter/gis_excerpt_converter.py
```diff
@@ -249,11 +262,13 @@ def execute_task(extraction_order_id, supported_export_formats, execution_config
...
                 raise
             finally:
                 try:
                     subprocess.check_call("docker-compose stop --timeout 0".split(' '))
                     subprocess.check_call("docker-compose rm -v -f".split(' '))
                 except:
+                    logger.exception("couldn't clean up docker containers")
```
wasabideveloper: Please pass `docker ps -a` too. May help to find the problem. Maybe it's a good idea to pass `docker-compose logs` too.

osmaxx-py/excerptconverter/dummyexcerptconverter/dummy_excerpt_converter.py
njordan-hsr: I like the message, thanks!

osmaxx-py/excerptconverter/gisexcerptconverter/gis_excerpt_converter.py
njordan-hsr: Python docs state:
```
Logger.exception(msg, *args, **kwargs)
Logs a message with level ERROR on this logger. The arguments are interpreted as for debug(). Exception info is added to the logging message. This method should only be called from an exception handler.
```

Doesn't that mean, I don't have to add any information, because it will be added to the log-message anyway?

osmaxx-py/excerptconverter/dummyexcerptconverter/dummy_excerpt_converter.py
njordan-hsr: [Python docs](https://docs.python.org/3.4/library/logging.html#logging.Logger.exception) state:
```
Logger.exception(msg, *args, **kwargs)
Logs a message with level ERROR on this logger. The arguments are interpreted as for debug().
Exception info is added to the logging message. This method should only be called from an exception handler.
```

Doesn't that mean, I don't have to add any information, because it will be added to the log-message anyway?

osmaxx-py/excerptconverter/gisexcerptconverter/gis_excerpt_converter.py
njordan-hsr: running `docker ps -a` is a really bad idea, because on a server, you might have hundreds of containers and a huge log for nothing. Furthermore, if this exception occurs, calling something with docker is a bad idea in general, because it might lead to the next exception. That's even more true for the `docker-compose logs`.
What we might add though, is the directory which it was run in - if it didn't get removed, we can look there (but it will be removed if nothing goes terribly wrong since we're using a 'tmp dir' in a with statement).

osmaxx-py/excerptconverter/dummyexcerptconverter/dummy_excerpt_converter.py
wasabideveloper: Yes

osmaxx-py/excerptconverter/gisexcerptconverter/gis_excerpt_converter.py
wasabideveloper: exception() will add stack trace and information about the exception by itself. You just need to pass a message.

osmaxx-py/excerptconverter/gisexcerptconverter/gis_excerpt_converter.py
wasabideveloper: I think it's really important to see the container status at the moment it failed. I agree `docker-compose ps` is better than `docker ps -a`. I think we should try to get this information, catch every errors may occour and pass the information we got.

osmaxx-py/excerptconverter/dummyexcerptconverter/dummy_excerpt_converter.py
```diff
@@ -89,10 +94,18 @@ def execute_task(extraction_order_id, supported_export_formats, execution_config
                 time.sleep(5)
                 wait_time += 5
                 if wait_time > 30:
+                    logger.exception(
+                        "Even after waiting for #{wait_time} #{time_unit_abbreviation}, the extraction order "
+                        "#{extraction_order_id} hasn't been created".format(
```
das-g: Uh, did you test this? The use of `#{...}` in [my comment](https://github.com/geometalab/osmaxx/pull/208#discussion_r40444228) was Ruby-inspired. It seems like [Python's `str.format()`](https://docs.python.org/3.4/library/stdtypes.html#str.format) uses [curly braces without the `#`](https://docs.python.org/3.4/library/string.html#formatstrings).

osmaxx-py/excerptconverter/gisexcerptconverter/gis_excerpt_converter.py
```diff
@@ -173,6 +181,14 @@ def execute_task(extraction_order_id, supported_export_formats, execution_config
                 time.sleep(5)
                 wait_time += 5
                 if wait_time > 30:
+                    logger.exception(
+                        "Even after waiting for #{wait_time} #{time_unit_abbreviation}, the extraction order "
+                        "#{extraction_order_id} hasn't been created".format(
```
das-g: Uh, did you test this? The use of `#{...}` in [my comment](https://github.com/geometalab/osmaxx/pull/208#discussion_r40445305) was Ruby-inspired. It seems like [Python's `str.format()`](https://docs.python.org/3.4/library/stdtypes.html#str.format) uses [curly braces without the `#`](https://docs.python.org/3.4/library/string.html#formatstrings).

osmaxx-py/excerptconverter/gisexcerptconverter/gis_excerpt_converter.py
```diff
@@ -255,5 +285,8 @@ def execute_task(extraction_order_id, supported_export_formats, execution_config
                     subprocess.check_call("docker-compose stop --timeout 0".split(' '))
                     subprocess.check_call("docker-compose rm -v -f".split(' '))
                 except:
+                    logger.exception("couldn't clean up docker containers in directory {directory}".format(
+                        directory=tmp_dir
+                    ))
                     pass
```
das-g: `pass` only makes sense for otherwise empty clauses. If a statement becomes non-empty, the `pass` should IMHO be removed. (Can `flake8` catch these?)

osmaxx-py/excerptconverter/dummyexcerptconverter/dummy_excerpt_converter.py
```diff
@@ -13,6 +14,9 @@
 from osmaxx.excerptexport import models
 from osmaxx.utils import private_storage
 
+# Get an instance of a logger
```
das-g: Is this explanation needed to understand the following line of code? What would you expect that line to do, were this comment not around?

If the comment turns out to be redundant, remove it.

osmaxx-py/excerptconverter/gisexcerptconverter/gis_excerpt_converter.py
```diff
@@ -16,6 +17,9 @@
 from osmaxx.excerptexport import models
 from osmaxx.utils import private_storage
 
+# Get an instance of a logger
```
das-g: If this comment turns out to be redundant, remove it.

osmaxx-py/excerptconverter/gisexcerptconverter/gis_excerpt_converter.py
```diff
@@ -115,21 +119,25 @@ def extract_excerpts(execution_configuration, extraction_order, bbox_args, conve
                                 email=False
                             )
                         else:
+                            message = _('The extraction of "%(file)s" of extraction order "%(order_id)s" failed.') % {
+                                'file': result_file_name,
+                                'order_id': extraction_order.id
+                            }
```
das-g: Do we have any rule-of-thumb when to use `%` and when `str.format`?

osmaxx-py/excerptconverter/dummyexcerptconverter/dummy_excerpt_converter.py
njordan-hsr: Uh, oh, it works, but prints `#` in there - thanks for the hint.

osmaxx-py/excerptconverter/gisexcerptconverter/gis_excerpt_converter.py
njordan-hsr: Pass is in this case indeed redundant, since pass is a noop and only needed for the compiler not to say anything if a statement is required.

osmaxx-py/excerptconverter/gisexcerptconverter/gis_excerpt_converter.py
njordan-hsr: yes: we generally went over to using `.format`, and older code is still using `%`. Even though I move the code only around, I think  it is a good idea to use `.format` more consistently, thanks for the input.

njordan-hsr:
After the last update, ran tests again with RUN_E2E=true ./test.sh, all green.

das-g:
:+1:

PR #209, opened 2015-09-27
Feature/206 move directory creation to storage

### LINKED ISSUE #206: Move directory creation to storage
Move the lines 150-151 of GisExcerptConverter:
```python
if not os.path.exists(private_storage.location):
    os.makedirs(private_storage.location)
```
to the storage itselfs and write a test.

njordan-hsr:
closes #206 

* small fix for the datetime warning during tests (use timezone aware `now()`)
* new class for easier testing of PrivateStorage
* tests for new class

ran tests with `RUN_E2E=true ./test.sh`, all green.

reviewed by:

* [ ] @das-g 
* [x] @wasabideveloper


osmaxx-py/osmaxx/tests/test_utils.py
```diff
@@ -0,0 +1,31 @@
...
+from django.test import TestCase
+import os
+from osmaxx.utils import PrivateSystemStorage
+
+
+class PrivateSystemStorageTestCase(TestCase):
+    def setUp(self):
+        random_directory_name_length = 10
+        self.random_directory_name = ''.join(
+            random.choice(string.ascii_lowercase + string.digits) for _ in range(random_directory_name_length)
```
wasabideveloper: What's this _ doing? I don't like not self-explaining variable names.

njordan-hsr: It signifies that this is a throwaway variable, it isn't used at all, just to be thrown away. It is quite common for Python to do it this way.

wasabideveloper: _ is used for translation too. I don't like this mess. Proposal: `_character`

osmaxx-py/osmaxx/tests/test_utils.py
```diff
@@ -0,0 +1,31 @@
+import string
+import random
+from django.test import TestCase
+import os
```
das-g: Don't we put standard library imports (`os`) before framework library imports (`django.test`)?

osmaxx-py/osmaxx/tests/test_utils.py
```diff
@@ -0,0 +1,31 @@
...
+
+class PrivateSystemStorageTestCase(TestCase):
+    def setUp(self):
+        random_directory_name_length = 10
+        self.random_directory_name = ''.join(
+            random.choice(string.ascii_lowercase + string.digits) for _ in range(random_directory_name_length)
+        )
+
+    def test_creates_a_new_directory_if_it_does_not_exist(self):
+        random_directory_path = os.path.join('/tmp/', self.random_directory_name)
```
das-g: What's the benefit of testing this with a random directory name instead of a hard-coded directory name? I'd get if you want to discover unforeseen edge cases and boundary conditions, but a lower-case alphanumeric name is unlikely to hit those, however random it might be.

osmaxx-py/osmaxx/tests/test_utils.py
njordan-hsr: good point, didn't think of that. Since this is not a character, but a number, can we settle for `_throwaway`?

osmaxx-py/osmaxx/tests/test_utils.py
njordan-hsr: It's more fun this way, there's no really reason behind it. Maybe choosing a `sensible`[1] is a better solution. Will change this.

[1]: one that doesn't clash with a possibly already existing directory under `/tmp/`

njordan-hsr:
After the last update, ran tests again with RUN_E2E=true ./test.sh, all green.

osmaxx-py/osmaxx/tests/test_utils.py
das-g: Yeah, I think I'd prefer an unlikely-to-clash sensible.

The book 'Clean Code' lists
> Don't be arbitrary

as a smell/heuristic (Chapter 17, Section 'General', **G32**).

njordan-hsr:
After the last update, ran tests again with `RUN_E2E=true ./test.sh`, all green.

osmaxx-py/osmaxx/tests/test_utils.py
wasabideveloper: :+1:

njordan-hsr:
`RUN_E2E=true ./test.sh` all green after merge (d77f9d3).

PR #210, opened 2015-09-27
Feature/192 handle removal of application containers in setup.production.sh

### LINKED ISSUE #192: Handle removal of application containers in setup.production.sh
Only remove application containers. Do not remove the data container.

Estimated duration: ~0.5d.

njordan-hsr:
Closes #192 

* delete all containers _except_ the data container.

Ran tests with `RUN_E2E=true ./test.sh`: all green.

Reviewed by:
* [ ] @das-g 
* [x] @wasabideveloper

### DIFF
--- a/compose-production.yml
+++ b/compose-production.yml
@@ -1,3 +1,5 @@
+# INFO: remember to update the setup.production.sh if you change something here.
+
 shareddata:
   image: buildpack-deps:jessie
   command: /bin/true
--- a/setup.production.sh
+++ b/setup.production.sh
@@ -10,16 +10,20 @@ docker create --name osmaxx-starter -v "$(pwd):/app"  -v "/var/run/docker.sock:/
 
 # stop containers & cleanup
 docker run -v "$(pwd):/app" -v "/var/run/docker.sock:/var/run/docker.sock" -e "COMPOSE_PROJECT_NAME=osmaxx" --rm "dduportal/docker-compose:$DOCKER_COMPOSE_TAG" stop
-# FIXME: data containers should not be removed!
-docker run -v "$(pwd):/app" -v "/var/run/docker.sock:/var/run/docker.sock" -e "COMPOSE_PROJECT_NAME=osmaxx" --rm "dduportal/docker-compose:$DOCKER_COMPOSE_TAG" rm -f
+# remove all containers explicitly except the data container
+docker run -v "$(pwd):/app" -v "/var/run/docker.sock:/var/run/docker.sock" -e "COMPOSE_PROJECT_NAME=osmaxx" --rm "dduportal/docker-compose:$DOCKER_COMPOSE_TAG" \
+rm -f webapp rabbitmq celery database email
 
 # pull & build
 docker run -v "$(pwd):/app" -v "/var/run/docker.sock:/var/run/docker.sock" -e "COMPOSE_PROJECT_NAME=osmaxx" --rm "dduportal/docker-compose:$DOCKER_COMPOSE_TAG" pull
-docker run -v "$(pwd):/app" -v "/var/run/docker.sock:/var/run/docker.sock" -e "COMPOSE_PROJECT_NAME=osmaxx" --rm "dduportal/docker-compose:$DOCKER_COMPOSE_TAG" build
+# build all containers explicitly except the data container
+docker run -v "$(pwd):/app" -v "/var/run/docker.sock:/var/run/docker.sock" -e "COMPOSE_PROJECT_NAME=osmaxx" --rm "dduportal/docker-compose:$DOCKER_COMPOSE_TAG" \
+build webapp rabbitmq celery database email
 
 # migrate
 docker run -v "$(pwd):/app" -v "/var/run/docker.sock:/var/run/docker.sock" -e "COMPOSE_PROJECT_NAME=osmaxx" --rm "dduportal/docker-compose:$DOCKER_COMPOSE_TAG" up -d database
 sleep 10
+# FIXME: create superuser should only be needed to be ran once.
 docker run -v "$(pwd):/app" -v "/var/run/docker.sock:/var/run/docker.sock" -e "COMPOSE_PROJECT_NAME=osmaxx" -it --rm "dduportal/docker-compose:$DOCKER_COMPOSE_TAG" \
     run --rm webapp /bin/bash -c "python3 manage.py migrate && python3 manage.py createsuperuser"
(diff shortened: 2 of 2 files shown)


njordan-hsr:
We might want to consider using a more robust way, because if the data container's base image gets updated, using pull will also trigger a rebuild.

Two options come to mind:

* [Option 1] Differentiating between bootstrapping and deploying:
    * Bootstrapping sets up all the containers (what the setup script did so far, including creating a superuser)
    * Deploying updates only the celery and the webapp container, and runs migrations, and keeps the rest as is.
* [Option 2] Never start/stop the database and data container, just boot them up (in theory you _should_ have to do this only once per deployment instance)

> In theory there is no difference between theory and practice. In practice there is.

wasabideveloper:
What about using tags for the container? If we refer a tag we are able to control the current used version and prevent rebuilds.

compose-production.yml
```diff
@@ -1,3 +1,5 @@
+# INFO: remember to update the setup.production.sh if you change something here.
```
wasabideveloper: Good hint :+1:

setup.production.sh
```diff
@@ -10,16 +10,20 @@ docker create --name osmaxx-starter -v "$(pwd):/app"  -v "/var/run/docker.sock:/
...
 docker run -v "$(pwd):/app" -v "/var/run/docker.sock:/var/run/docker.sock" -e "COMPOSE_PROJECT_NAME=osmaxx" --rm "dduportal/docker-compose:$DOCKER_COMPOSE_TAG" pull
-docker run -v "$(pwd):/app" -v "/var/run/docker.sock:/var/run/docker.sock" -e "COMPOSE_PROJECT_NAME=osmaxx" --rm "dduportal/docker-compose:$DOCKER_COMPOSE_TAG" build
+# build all containers explicitly except the data container
+docker run -v "$(pwd):/app" -v "/var/run/docker.sock:/var/run/docker.sock" -e "COMPOSE_PROJECT_NAME=osmaxx" --rm "dduportal/docker-compose:$DOCKER_COMPOSE_TAG" \
+build webapp rabbitmq celery database email
 
 # migrate
 docker run -v "$(pwd):/app" -v "/var/run/docker.sock:/var/run/docker.sock" -e "COMPOSE_PROJECT_NAME=osmaxx" --rm "dduportal/docker-compose:$DOCKER_COMPOSE_TAG" up -d database
 sleep 10
+# FIXME: create superuser should only be needed to be ran once.
```
wasabideveloper: It is ran only once. ???

setup.production.sh
```diff
@@ -10,16 +10,20 @@ docker create --name osmaxx-starter -v "$(pwd):/app"  -v "/var/run/docker.sock:/
 
 # stop containers & cleanup
 docker run -v "$(pwd):/app" -v "/var/run/docker.sock:/var/run/docker.sock" -e "COMPOSE_PROJECT_NAME=osmaxx" --rm "dduportal/docker-compose:$DOCKER_COMPOSE_TAG" stop
-# TODO: data containers should not be removed!
-docker run -v "$(pwd):/app" -v "/var/run/docker.sock:/var/run/docker.sock" -e "COMPOSE_PROJECT_NAME=osmaxx" --rm "dduportal/docker-compose:$DOCKER_COMPOSE_TAG" rm -f
+# remove all containers explicitly except the data container
+docker run -v "$(pwd):/app" -v "/var/run/docker.sock:/var/run/docker.sock" -e "COMPOSE_PROJECT_NAME=osmaxx" --rm "dduportal/docker-compose:$DOCKER_COMPOSE_TAG" \
+rm -f webapp rabbitmq celery database email
```
das-g: Would it make sense to have a variable for 'all containers except for the data container'?

setup.production.sh
```diff
@@ -10,16 +10,20 @@ docker create --name osmaxx-starter -v "$(pwd):/app"  -v "/var/run/docker.sock:/
...
+docker run -v "$(pwd):/app" -v "/var/run/docker.sock:/var/run/docker.sock" -e "COMPOSE_PROJECT_NAME=osmaxx" --rm "dduportal/docker-compose:$DOCKER_COMPOSE_TAG" \
+build webapp rabbitmq celery database email
 
 # migrate
 docker run -v "$(pwd):/app" -v "/var/run/docker.sock:/var/run/docker.sock" -e "COMPOSE_PROJECT_NAME=osmaxx" --rm "dduportal/docker-compose:$DOCKER_COMPOSE_TAG" up -d database
 sleep 10
+# FIXME: create superuser should only be needed to be ran once.
 docker run -v "$(pwd):/app" -v "/var/run/docker.sock:/var/run/docker.sock" -e "COMPOSE_PROJECT_NAME=osmaxx" -it --rm "dduportal/docker-compose:$DOCKER_COMPOSE_TAG" \
     run --rm webapp /bin/bash -c "python3 manage.py migrate && python3 manage.py createsuperuser"
 docker run -v "$(pwd):/app" -v "/var/run/docker.sock:/var/run/docker.sock" -e "COMPOSE_PROJECT_NAME=osmaxx" --rm "dduportal/docker-compose:$DOCKER_COMPOSE_TAG" stop
```
das-g: `docker run -v "$(pwd):/app" -v "/var/run/docker.sock:/var/run/docker.sock" -e "COMPOSE_PROJECT_NAME=osmaxx" --rm "dduportal/docker-compose:$DOCKER_COMPOSE_TAG"` sure occurs a lot in this file. Can we DRY this?

setup.production.sh
njordan-hsr: No, but once you have a database, you shouldn't be needing to run this every time you deploy!

wasabideveloper: :+1: We should check if the database already exits and if there are migrations to apply and ask the user if he would like to apply. But this out of scope. Create a new issue for.

setup.production.sh
njordan-hsr: good idea, though we really ought only to rebuild celery and webapp, all other (including rabbitmq and the database) should be left as is.

setup.production.sh
njordan-hsr: We can :)

njordan-hsr:
Using tags is a good idea, only that it is trickier: automated builds on Dockerhub no longer work...

setup.production.sh
njordan-hsr: IMO I don't think we should handle this case at all. If you deploy, deploy it normally, and handle creating a superuser separately (not part of a deployment script, but as information in the documentation only).

Alternatively, make a separation between bootstrapping and updating?

wasabideveloper: There is no bootstrapping in the script. It's only setup (apply migrations & create superuser).

wasabideveloper:
Why can't it be simple just once :wink:

njordan-hsr:
`RUN_E2E=true ./test.sh` green after merge with develop.

PR #212, opened 2015-09-29
Feature/test gis converter#187

### LINKED ISSUE #187: Write tests for GIS excerpt converter
Estimated duration: ~2d

wasabideveloper:
Implementation of feature #187 (Merge will close #187 ).

* ran tests
* tested views by hand:
  * /orders/new [get/post]
  * /orders/{id}

To be reviewed by:
- [x] @njordan-hsr 
- [ ] @das-g


wasabideveloper:
I fixed the failing export. It's ready for review and merge now.

osmaxx-py/excerptconverter/dummyexcerptconverter/dummy_excerpt_converter.py
```diff
@@ -7,120 +7,126 @@
...
-        while extraction_order is None:
-            try:
-                extraction_order = models.ExtractionOrder.objects.get(pk=extraction_order_id)
-            except models.ExtractionOrder.DoesNotExist:
-                time.sleep(5)
-                wait_time += 5
-                if wait_time > 30:
-                    raise
-
+name = 'Dummy'
```
das-g: `name`, `export_formats` and `export_options` should never be modified at runtime, so make them constants (i.e. give them ALL_CAPS names).

osmaxx-py/excerptconverter/dummyexcerptconverter/dummy_excerpt_converter.py
```diff
@@ -7,120 +7,126 @@
...
+export_formats = {
+    'txt': {
+        'name': 'Text (.txt)',
+        'file_extension': 'txt',
+        'mime_type': 'text/plain'
+    },
```
das-g: Would it make sense to have the dict values (representing a single format) as objects of a yet-to-be-created class instead of as dicts themselves? That way, if the required attributes change, not-yet-updated formats would already fail at construction rather than much later when being used.

@njordan-hsr would a class without methods be unpythonic?

osmaxx-py/excerptconverter/converter_helper.py
```diff
@@ -7,6 +6,50 @@
...
+def module_converter_configuration(name, export_formats, export_options):
+    return {
+        'name': name,
+        'formats': export_formats,
+        'options': export_options
+    }
+
+
+def run_model_execute(execute_task_method, export_formats,
+                      extraction_order, execution_configuration, run_as_celery_tasks):
```
das-g: Avoid positional boolean arguments. (See 'Clean Code', Chapter 17, Section 'Functions', **F3: Flag Arguments** and Section 'General', **G15: Selector Arguments**.) Splitting the function into two separate ones as 'Clean Code' recommends is not practical here, though.

Thus, can we use a keyword argument instead, with `True` as default value? That way, only the invocation in some tests will be cluttered, and the "clutter" will help understand the meaning of this flag.

osmaxx-py/excerptconverter/dummyexcerptconverter/dummy_excerpt_converter.py
njordan-hsr: @das-g no, why would it? It is quite common to have Mix-ins (multi-inheritance) for just that (but mostly for methods, that is true).

osmaxx-py/excerptconverter/dummyexcerptconverter/dummy_excerpt_converter.py
wasabideveloper: :+1:

osmaxx-py/excerptconverter/dummyexcerptconverter/dummy_excerpt_converter.py
wasabideveloper: I tested some variants but we won't gain something.
:see_no_evil:

osmaxx-py/excerptconverter/converter_helper.py
wasabideveloper: :heavy_check_mark:

osmaxx-py/excerptconverter/converter_helper.py
```diff
@@ -7,6 +6,42 @@
...
+                'default': '500'
+            },
+            'quality': {
+                'label': 'Quality',
+                'type': 'number',
+                'default': '10'
+            }
+        }
+    """
+    return {
```
njordan-hsr: does this warrant a function? Or is this specifically an addition needed to be overridden during testing?

osmaxx-py/excerptconverter/converter_registry.py
```diff
@@ -0,0 +1 @@
+available_converters = []
```
njordan-hsr: that's a very simple registry :+1:

osmaxx-py/excerptconverter/dummyexcerptconverter/dummy_excerpt_converter.py
njordan-hsr: Like this, they can be interpreted (by a programmer) as contsants, which I think is correct. If they should be on module level or class level (ie. mixin?) is, IMHO, not that important.

> Special cases aren't special enough to break the rules.
> Although practicality beats purity.

osmaxx-py/excerptconverter/dummyexcerptconverter/dummy_excerpt_converter.py
```diff
@@ -7,120 +7,116 @@
...
+        'default': 'verbatim',
+        'values': [
+            {'name': 'verbatim', 'label': 'Verbatim'},
+            {'name': 'simplified', 'label': 'Simplified'}
+        ]
+    }
+}
+
+
+def converter_configuration():
```
njordan-hsr: Another function to wrap another function without doing really anything, seems a bit repetitious to me - is it correct, that this doesn't get used at all (except in the tests)? Then I'd argue this'd be moved there.

osmaxx-py/excerptconverter/gisexcerptconverter/gis_excerpt_converter.py
```diff
@@ -4,256 +4,262 @@
...
-            #     'label': 'Detail level',
-            #     'type': 'choice',
-            #     'default': 'verbatim',
-            #     'values': [
-            #         {'name': 'verbatim', 'label': 'Verbatim'},
-            #         {'name': 'simplified', 'label': 'Simplified'}
-            #     ]
-            # }
-        }
+def converter_configuration():
```
njordan-hsr: Another function to wrap another function without doing really anything, seems a bit repetitious to me - is it correct, that this doesn't get used at all (except in the tests)? Then I'd argue this'd be moved there.

osmaxx-py/excerptconverter/converter_helper.py
das-g: > [...] is this specifically an addition needed to be overridden during testing?

No.
> does returning a dictionary from input variables without any changes warrant a function?

The function's signature enforces presence of expected components. Whether that benefit is enough to warrant a function, I'm not sure.

wasabideveloper: It's what das-g said.

osmaxx-py/excerptconverter/dummyexcerptconverter/dummy_excerpt_converter.py
wasabideveloper: We can't inherit this function because we are no longer using classes. So it's the best solution to implement the returned data structure ONLY once.
@njordan-hsr Do you have a better idea?

osmaxx-py/excerptconverter/gisexcerptconverter/gis_excerpt_converter.py
wasabideveloper: This function is used by the converter manager and so by the options form.

osmaxx-py/excerptconverter/dummyexcerptconverter/dummy_excerpt_converter.py
njordan-hsr: No argument there, and you should be able to import the data structure directly, no need for a function returning it.

osmaxx-py/excerptconverter/gisexcerptconverter/gis_excerpt_converter.py
njordan-hsr: I probably should have rephrased the question: Why do we need a function to return a dictionary?

osmaxx-py/excerptconverter/converter_helper.py
njordan-hsr: >The function's signature enforces presence of expected components

No, they don't, they still can be passed as `None` and everything looks fine. I'd argue it's even worse because an exception doesn't come up at the right place: since the dictionary has that item defined an is None it will raise some other exception at (probably quite unexpected) place, instead of raising a KeyError at the very instance of access.

I'd rather have a constant to make it clear, that this is a, well, (config) constant.

das-g: >>The function's signature enforces presence of expected components
> 
> No, they don't, they still can be passed as `None` and everything looks fine.

'enforces' might have been the wrong term here. What I meant, was: Using that function can guide us by its signature. If we later add another required component, we'll just expand the signature accordingly and all calls that need to up updated will be highlighted as errors in the IDE. Of course, a determined coder could still just not use the function or pass `None` or even nonsense to it.

osmaxx-py/excerptconverter/dummyexcerptconverter/dummy_excerpt_converter.py
wasabideveloper: I created an issue: #216 to enhance this

osmaxx-py/excerptconverter/gisexcerptconverter/gis_excerpt_converter.py
wasabideveloper: We should change this the way we discussed it: #216

osmaxx-py/excerptconverter/converter_helper.py
njordan-hsr: I can see the use of this, I guess. Let's just leave it as it is, but making this into a config (ie. settings) using something like an [ConfigParser](https://docs.python.org/3.4/library/configparser.html) is IMHO the better approach, if it really is a configuration (it does look like it ;) )

wasabideveloper: We will change this the way we discussed it: #216

njordan-hsr:
From my perspective, we can merge this.

osmaxx-py/excerptconverter/dummyexcerptconverter/dummy_excerpt_converter.py
wasabideveloper: As proposed on #216 we should verify the correct format by a test. I think that's enough.

wasabideveloper: ( :heavy_check_mark: )

osmaxx-py/excerptconverter/converter_helper.py
wasabideveloper: ( :heavy_check_mark: )

osmaxx-py/excerptconverter/dummyexcerptconverter/dummy_excerpt_converter.py
wasabideveloper: ( :checkered_flag: )

osmaxx-py/excerptconverter/gisexcerptconverter/gis_excerpt_converter.py
wasabideveloper: ( :heavy_check_mark: )

wasabideveloper:
@das-g I agree

PR #213, opened 2015-09-30
feature/speedup container building for development

njordan-hsr:
Closes no issue at all, but helps nonetheless ;-)

Using a little trick the tests to speedup the build process.

`RUN_E2E=true ./test.sh` 

Reviewed by:

* [x] @das-g 
* [ ] @wasabideveloper

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,8 @@ ENV HOME /home/$USER
 
 WORKDIR $HOME
 
-# if you update your requirements, please update this to the actual date/time!
+# if you update your requirements, please update this to the actual date/time, 
+# otherwise docker uses the cache from the intermediate image build (not re-running pip3).
 ENV REQS_LAST_UPDATED 30-09-2015 11:48
 
 ADD osmaxx-py/requirements $HOME/requirements
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,10 +12,17 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get update -q \
 
 ENV HOME /home/$USER
 
+WORKDIR $HOME
+
+# little trick to avoid rebuilding this image over and over again, if you update
+# your requirements, please update this to the actual date!
+
+ENV REQS_LAST_UPDATED 30-09-2015 11:48
+
+ADD osmaxx-py/requirements $HOME/requirements
+
+RUN pip3 install -r requirements/local.txt
+
 WORKDIR $HOME/source
 
 COPY osmaxx-py $HOME/source
-
-ENV REQS_LAST_UPDATED 22-06-2015 14:24
-
-RUN pip3 install -Ur requirements/local.txt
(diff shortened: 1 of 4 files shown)


test.sh
```diff
@@ -99,7 +99,7 @@ function tear_down() {
 
 function reset_containers() {
     docker_compose stop -t 0 &>> ${LOGFILE};
-    docker_compose rm -vf &>> ${LOGFILE};
+    docker_compose rm -f &>> ${LOGFILE};
```
das-g: If we don't rely on volumes for the caching, this change isn't needed for the stated purpose of this pull request. (Maybe it's a good idea anyway, though.)

test.sh
```diff
@@ -108,7 +108,7 @@ function reset_containers() {
 function reset_container() {
     local CONTAINER_TO_BE_RESETTED=$1
     docker_compose stop -t 0 ${CONTAINER_TO_BE_RESETTED} &>> ${LOGFILE};
-    docker_compose rm -vf ${CONTAINER_TO_BE_RESETTED} &>> ${LOGFILE};
+    docker_compose rm -f ${CONTAINER_TO_BE_RESETTED} &>> ${LOGFILE};
```
das-g: If we don't rely on volumes for the caching, this change isn't needed for the stated purpose of this pull request. (Maybe it's a good idea anyway, though.)

e2e/helpers/docker_compose.py
```diff
@@ -10,7 +10,7 @@ def stop():
 
 def clean():
     stop()
-    subprocess.check_call("docker-compose rm -vf".split(' '))
+    subprocess.check_call("docker-compose rm -f".split(' '))
```
das-g: If we don't rely on volumes for the caching, this change isn't needed for the stated purpose of this pull request. (Maybe it's a good idea anyway, though.)

Dockerfile
```diff
@@ -12,10 +12,17 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get update -q \
 
 ENV HOME /home/$USER
 
-WORKDIR $HOME/source
+WORKDIR $HOME
 
-COPY osmaxx-py $HOME/source
+# Little trick to avoid rebuilding this image over and over again.
```
das-g: Instead of stating that this is a 'trick' (whatever that means), this comment should explain or highlight what the trick is and how it works.

njordan-hsr: oh, so it isn't obvious... probably missing it, because I didn't know, what to write as an explanation; should probably just remove it, since it doesn't add anything useful.

test.sh
njordan-hsr: That's true, and maybe it's even a good idea to leave it in, since it assures a clean state. I will revert this change.

njordan-hsr:
ran `RUN_E2E=true ./test.sh` after merge and update (4f9ad91), all green.

Dockerfile
wasabideveloper: I would like to see a comment why it's better to run pip first. I know, our comments should not explain how docker works. But in reality it helps to avoid bad refactorings.

http://www.codecandies.com/2011/04/13/please-increment-the-following-counter/

Issue #189, opened 2015-09-14
Fix email shipping


wasabideveloper:
There are no emails sent yet on successful extraction.

Estimated duration: ~1d

njordan-hsr:
This is also a follow-up on issue #191 , because it depends on how we resolve that issue there.

njordan-hsr:
As soon as #214 is merged, we can start re-enabling sending emails and adding a test to resolve this bug.

PR #214, opened 2015-09-30
Feature/191 remove email credentials from docker compose

### LINKED ISSUE #191: Remove email credentials from docker compose
Login credentials should not be stored in docker compose and published on github.

Estimated duration: ~0.5d

### LINKED ISSUE #193: Fix failing removal if temporary containers of excerpt container chain

njordan-hsr:
Closes #191 

Ran test: `RUN_E2E=true ./test.sh`: all green.

Reviewed by:

* [ ] @das-g 
* [ ] @wasabideveloper

### DIFF
--- a/compose-development.yml
+++ b/compose-development.yml
@@ -42,16 +42,6 @@ webappdev:
     - DJANGO_DEBUG=true
     - DJANGO_EMAIL_FILE_PATH=/dev-emails
     - DJANGO_EMAIL_BACKEND=django.core.mail.backends.filebased.EmailBackend
-    - DJANGO_EMAIL_HOST=localhost
-    - DJANGO_EMAIL_HOST_PASSWORD=''
-    - DJANGO_EMAIL_HOST_USER=''
-    - DJANGO_EMAIL_PORT=25
-    - DJANGO_EMAIL_SUBJECT_PREFIX='[OSMAXX-DEV] '
-    - DJANGO_EMAIL_USE_TLS=False
-    - DJANGO_EMAIL_USE_SSL=False
-    - DJANGO_EMAIL_TIMEOUT=None
-    - DJANGO_EMAIL_SSL_CERTFILET=None
-    - DJANGO_EMAIL_SSL_KEYFILE=None
 
 celerydev:
   extends:
@@ -72,16 +62,6 @@ celerydev:
     - DJANGO_DEBUG=true
     - DJANGO_EMAIL_FILE_PATH=/dev-emails
     - DJANGO_EMAIL_BACKEND=django.core.mail.backends.filebased.EmailBackend
-    - DJANGO_EMAIL_HOST=localhost
-    - DJANGO_EMAIL_HOST_PASSWORD=''
-    - DJANGO_EMAIL_HOST_USER=''
-    - DJANGO_EMAIL_PORT=25
-    - DJANGO_EMAIL_SUBJECT_PREFIX='[OSMAXX-DEV] '
-    - DJANGO_EMAIL_USE_TLS=False
-    - DJANGO_EMAIL_USE_SSL=False
-    - DJANGO_EMAIL_TIMEOUT=None
-    - DJANGO_EMAIL_SSL_CERTFILET=None
-    - DJANGO_EMAIL_SSL_KEYFILE=None
 
 rabbitmqdev:
   extends:
--- a/compose-development.yml
+++ b/compose-development.yml
@@ -10,6 +10,8 @@ shareddatadev:
     - /database/data
     # temporary results directory used as pipe between multiple docker forests
     - /tmp/osmaxx-results:/results
+    # emails file from email sending during development and testing
+    - /tmp/osmaxx-development-emails:/dev-emails
   net: "none"
 
 sourcedev:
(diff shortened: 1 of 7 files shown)


das-g:
Are you sure this is related to #193, (which already has been closed by a different pull request)?

das-g:
Probably should refer to #189 and/or #191.

compose-development.yml
```diff
@@ -33,19 +35,23 @@ webappdev:
...
-rabbitmqdev:
-  extends:
-    file: compose-common.yml
-    service: rabbitmqbase
+    - DJANGO_EMAIL_FILE_PATH=/dev-emails
+    - DJANGO_EMAIL_BACKEND=django.core.mail.backends.filebased.EmailBackend
+    - DJANGO_EMAIL_HOST=localhost
+    - DJANGO_EMAIL_HOST_PASSWORD=''
+    - DJANGO_EMAIL_HOST_USER=''
+    - DJANGO_EMAIL_PORT=25
```
das-g: Do we need all of `DJANGO_EMAIL_HOST`, `DJANGO_EMAIL_HOST_PASSWORD`, `DJANGO_EMAIL_HOST_USER`, `DJANGO_EMAIL_PORT` when using `DJANGO_EMAIL_BACKEND=django.core.mail.backends.filebased.EmailBackend`?

compose-production.yml
```diff
@@ -21,18 +21,21 @@ webapp:
...
-    - DJANGO_EMAIL_PORT=25
+    - DJANGO_EMAIL_BACKEND=django.core.mail.backends.smtp.EmailBackend
+    - DJANGO_EMAIL_HOST=<changeme>
+    - DJANGO_EMAIL_HOST_PASSWORD='<changeme>'
+    - DJANGO_EMAIL_HOST_USER='<changeme>'
+    - DJANGO_EMAIL_PORT=<changeme>
+    - DJANGO_EMAIL_SUBJECT_PREFIX='<changeme>'
+    - DJANGO_EMAIL_USE_TLS=<changeme>
+    - DJANGO_EMAIL_USE_SSL=<changeme>
+    - DJANGO_EMAIL_TIMEOUT=<changeme>
```
das-g: Why set ENV variables that need to be overwritten anyway?

docs/deployment.md
```diff
@@ -14,7 +14,21 @@ To deploy to production server:
   git submodule init && git submodule update
   ```
   Repository details see [Project repository](git-repository.md).
-3. Link production configuration for docker-compose, Details see [Docker container bootstrapping](../README.md#initializationdocker-container-bootstrapping).
+3. a. Link production configuration for docker-compose, Details see [Docker container bootstrapping](../README.md#initializationdocker-container-bootstrapping).
+   b. Edit the docker-compose.yml file and change (if needed) the EMAIL settings:
+   ```shell
+    DJANGO_EMAIL_HOST='<your smtp>'
```
das-g: This would be a domain name, I guess? An example (e.g. using `smtp.example.com`) might be more useful than a placeholder. If an explanation is needed in addition to the example, append a shell comment (`# ...`).

docs/deployment.md
```diff
@@ -14,7 +14,21 @@ To deploy to production server:
...
+    DJANGO_EMAIL_HOST_PASSWORD='<your password>'
+    DJANGO_EMAIL_HOST_USER='<your user>'
+    DJANGO_EMAIL_PORT=<PORT>
+    DJANGO_EMAIL_SUBJECT_PREFIX='<set to something smart>'
+    DJANGO_EMAIL_USE_TLS=<true/false>
```
das-g: is this case sensitive?

docs/deployment.md
```diff
@@ -14,7 +14,21 @@ To deploy to production server:
...
+    DJANGO_EMAIL_USE_SSL=<true/false>
+    DJANGO_EMAIL_TIMEOUT=<None or Number>
```
das-g: Would the number be in &mu;s, ms, s, minutes, hours, days, ...?
I guess `None` means **never** time out. Will `0` be treated the same as `None` or cause an instant timeout?

docs/deployment.md
njordan-hsr: You mean the true/false? [No](https://github.com/joke2k/django-environ/blob/develop/environ/environ.py#L284).

docs/deployment.md
njordan-hsr: It's the default setting in Django. Have a look at the [documentation for this setting](https://docs.djangoproject.com/en/1.8/ref/settings/#email-timeout).

compose-development.yml
njordan-hsr: No, but it is easier to remember that those are the settings, if you test it out in development with a more real scenario. We can remove it or comment the unused out as other possibilities. Any preference?

compose-production.yml
njordan-hsr: Otherwise you might forget to set them properly and like this, the project doesn't run, if they are not set correctly. And if they don't have placeholders, it wouldn't be that self-explaining anymore (are we being to explicit here?).

docs/deployment.md
njordan-hsr: good idea, didn't think of that.

compose-development.yml
das-g: I'd prefer removing them.

compose-production.yml
das-g: Would the project not run at all, or would it fail upon trying to send the first email? If the latter, I see no benefit in overriding here the (reasonably sane) defaults you've put into `osmaxx-py/config/settings/*.py`, only to force the operator to again overwrite them.

njordan-hsr: You're right, it would run, even docker compose would start. Wrong thinking on my end, I'll change that to documentation and simple configuration.

njordan-hsr:
Ran `RUN_E2E=true ./test.sh`: all green after last update (5b191ec).

docs/deployment.md
```diff
@@ -14,7 +14,23 @@ To deploy to production server:
   git submodule init && git submodule update
   ```
   Repository details see [Project repository](git-repository.md).
-3. Link production configuration for docker-compose, Details see [Docker container bootstrapping](../README.md#initializationdocker-container-bootstrapping).
+3. a. Link production configuration for docker-compose, Details see [Docker container bootstrapping](../README.md#initializationdocker-container-bootstrapping).
+   b. Edit the `compose-production.yml` file and change/add environment settings for celery and webapp. 
+      List of available additional settings (no quotes for strings!):
+   
+       ```
+        - DJANGO_EMAIL_HOST=<your smtp>   # ie. smtp.example.com
```
das-g: * You mean 'e.g.' (for example), not 'i.e.' (it is), don't you?
* I'd replace the placeholder by the example and move the explanation within the placeholder to the comment:

        - DJANGO_EMAIL_HOST=smtp.example.com   # The SMTP server to use for sending mails to users

    That way, you avoid ambiguity in the syntax, e.g. someone thinking they'd have to use

        - DJANGO_EMAIL_HOST=<mail.my.example.com>
    with literal `<` and `>` when `mail.my.example.com` is their server's domain.

njordan-hsr:
LGTM (let's get this merged and looks good to me ;-) )

njordan-hsr:
@das-g @wasabideveloper still all tests green after (7c40b19), can we merge it?

setup.production.sh
```diff
@@ -1,5 +1,18 @@
 #!/bin/bash
 
+# Asking is better then forgetting
+function ask_if_docker_compose_file_has_been_updated() {
+    echo "Are the settings in the compose-production.yml correct and is compose-production.yml available through docker-compose.yml (symlink it)?"
```
das-g: The script itself could probably check the symlink instead of asking the user.

njordan-hsr: I guess it could, but then it would not be possible anymore to copy it, which sometimes might be what the user would want.

das-g: What if we check by using `diff`?

PR #217, opened 2015-10-01
Feature/189 fix email shipping

### LINKED ISSUE #189: Fix email shipping

njordan-hsr:
Closes #189 

Ran tests with `RUN_E2E=true ./test.sh`, all green.

Reviewd by:

* [ ] @das-g 
* [x] @wasabideveloper

### DIFF
--- a/osmaxx-py/excerptconverter/converter_helper.py
+++ b/osmaxx-py/excerptconverter/converter_helper.py
@@ -1,3 +1,4 @@
+from django.conf import settings
 from django.contrib import messages
 from django.core import mail
 from django.utils.translation import ugettext_lazy as _
@@ -74,7 +75,7 @@ class ConverterHelper:
                 mail.send_mail(
                     '[OSMAXX] '+message_text,
                     message_text,
-                    'no-reply@osmaxx.hsr.ch',
+                    settings.DEFAULT_FROM_EMAIL,
                     [self.user.email]
                 )
             else:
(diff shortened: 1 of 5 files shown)


wasabideveloper:
@njordan-hsr You didn't test the email shipping out of the test scope (=real shipping), right?

njordan-hsr:
What do you mean with real shipping? `send_mail` is a Django function, which shouldn't be needed to be tested on our side, so unless we're doing something wrong in our settings, this does work.

This doesn't work 'in reality' since it need the changes from #214 to be able to send mails correctly.

I merged that locally and ran it manually: it sends emails to the console, also using some valid smtp settings, I got a result email, if that is what you're after.

Ran `RUN_E2E=true ./test.sh` after last commit (b43979d), all green.

wasabideveloper:
I don't want to test django `send_mail` but I would like to see the whole setup working (e.g. routing the email out of the docker forest...). But it's ok. I will merge it an we will test it after.
:heavy_check_mark:

osmaxx-py/excerptconverter/converter_helper.py
```diff
@@ -70,14 +72,12 @@ def inform_user(self, message_type, message_text, email=True):
...
-                pass
-                # FIXME: fix email sending
-                # send_mail(
-                #     '[OSMAXX] '+message_text,
-                #     message_text,
-                #     'no-reply@osmaxx.hsr.ch',
-                #     [self.user.email]
-                # )
+                mail.send_mail(
+                    '[OSMAXX] '+message_text,
```
das-g: Isn't the hardcoded prefix here redundant to the `EMAIL_SUBJECT_PREFIX` environment variable?

PR #226, opened 2015-10-04
Feature/readying for deployment

njordan-hsr:
Many many changes, sorry about that...

* correct wrong settings for celery
* better default settings in compose-common
* include more settings, which should be configurable
* added ssl settings as comments
* added and removed again: deploy script

Tests:
* Ran tests with `RUN_E2E=true ./test.sh`: all green.
* Ran `./setup.production.sh`:
    * if ran more than once, it keeps the database: true
* allowing login without csrf errors: true
* login with clavid succeeds: true
* extraction order can be placed as admin and non-admin: true

Reviewed by:
* [x]  @das-g


compose-common.yml
```diff
@@ -12,35 +12,49 @@ base:
     # --link some-rabbit:rabbit "just works"
     - DJANGO_CELERY_BROKER_URL=amqp://guest@rabbit
     - DJANGO_DATABASE_URL=postgis://postgres@database/postgres
-    - DJANGO_EMAIL_HOST=localhost
-    # Settings above likely need to be changed, depending on  the environment
+    - APP_PORT=8000
+    - APP_HOST=0.0.0.0
+
+    # Logging
+    - PYTHONUNBUFFERED=non-empty-string
```
das-g: '`non-empty-string`'? What's going on here?

The famous WTFs/m cartoon

compose-common.yml
```diff
@@ -12,35 +12,49 @@ base:
...
+    - PYTHONIOENCODING=utf-8
+    - DJANGO_LOG_LEVEL=INFO
+    # https://docs.djangoproject.com/en/1.8/howto/deployment/checklist/#python-options
+    - PYTHONHASHSEED=random
+
+
+    # Settings only need to be changed, if docker-compose is being changed
+    # or not used at all.
     ##########################################################################
```
das-g: I guess this applies for settings below this comment?

compose-common.yml
```diff
@@ -12,35 +12,49 @@ base:
     # --link some-rabbit:rabbit "just works"
     - DJANGO_CELERY_BROKER_URL=amqp://guest@rabbit
     - DJANGO_DATABASE_URL=postgis://postgres@database/postgres
-    - DJANGO_EMAIL_HOST=localhost
-    # Settings above likely need to be changed, depending on  the environment
+    - APP_PORT=8000
+    - APP_HOST=0.0.0.0
```
das-g: `0.0.0.0` is for exposing on all network interfaces, isn't it?

Maybe warrants a comment.

compose-common.yml
```diff
@@ -12,35 +12,51 @@ base:
...
-    - DJANGO_SECRET_KEY=`< /dev/urandom tr -dc _A-Z-a-z-0-9 | head -c120`
-    - DJANGO_EMAIL_BACKEND=django.core.mail.backends.smtp.EmailBackend
-    # EMAIL
-    - DJANGO_DEFAULT_FROM_EMAIL='noreply osmaxx <noreply@osmaxx.hsr.ch>'
 
-    # SECURITY
-    - DJANGO_ALLOWED_HOSTS='[".osmaxx.hsr.ch.",]'
+    # port and host application listens
+    - APP_PORT=8000
+    - APP_HOST=0.0.0.0
```
das-g: These were already set [above](https://github.com/geometalab/osmaxx/pull/226/files#diff-c21c236fe2cc3f26f9abcd1be45b837bR15).

compose-common.yml
```diff
@@ -12,35 +12,51 @@ base:
...
-    - DJANGO_SECURE_HSTS_SECONDS=60
-    - DJANGO_SECURE_HSTS_INCLUDE_SUBDOMAINS=true
-    - DJANGO_SECURE_SSL_REDIRECT=true
-    - DJANGO_SECURE_SSL_HOST='osmaxx.hsr.ch'
-    - DJANGO_SECURE_REDIRECT_EXEMPT=[]
+    # EMAIL
+    - DJANGO_EMAIL_BACKEND=django.core.mail.backends.smtp.EmailBackend
+    - DJANGO_EMAIL_HOST=localhost
+    - DJANGO_EMAIL_HOST_PASSWORD=
+    - DJANGO_EMAIL_HOST_USER=
```
das-g: Are password and user being set to empty strings?

osmaxx-py/config/settings/common.py
```diff
@@ -98,13 +98,19 @@
 EMAIL_HOST_PASSWORD = env.str('DJANGO_EMAIL_HOST_PASSWORD', default='')
 EMAIL_HOST_USER = env.str('DJANGO_EMAIL_HOST_USER', default='')
 EMAIL_PORT = env.int('DJANGO_EMAIL_PORT', default=25)
-EMAIL_SUBJECT_PREFIX = env.str('DJANGO_EMAIL_SUBJECT_PREFIX', default='')
+EMAIL_SUBJECT_PREFIX = env.str('DJANGO_EMAIL_SUBJECT_PREFIX', default='[Django] ')
```
das-g: Wouldn't we want `'[Osmaxx]'`, here?

osmaxx-py/config/settings/common.py
```diff
@@ -388,3 +381,44 @@
...
+# Celery
+BROKER_URL = env.str('DJANGO_CELERY_BROKER_URL', default='amqp://guest:guest@localhost:5672//')
+CELERY_RESULT_BACKEND = BROKER_URL
+
+# Security - defaults taken from Django 1.8
+SECRET_KEY = env.str("DJANGO_SECRET_KEY", default=None)
+ALLOWED_HOSTS = env.list('DJANGO_ALLOWED_HOSTS', default=[])
+X_FRAME_OPTIONS = env.str('DJANGO_X_FRAME_OPTIONS', default='SAMEORIGIN')
+
+## General SECURE Settings - defaults are not secured!
```
das-g: What defaults are not secure(d)? Django's defaults? The defaults we set here with `default=`?

compose-common.yml
```diff
@@ -12,35 +12,51 @@ base:
...
+    - DJANGO_EMAIL_USE_SSL=false
+    - DJANGO_DEFAULT_FROM_EMAIL=no-reply@osmaxx.dev
+    - DJANGO_SERVER_EMAIL=no-reply@osmaxx.dev
 
-    # OTHER SECURITY SETTINGS
-    - DJANGO_SECURE_CONTENT_TYPE_NOSNIFF=true
-    - DJANGO_SECURE_BROWSER_XSS_FILTER=true
-    - DJANGO_SESSION_COOKIE_SECURE=true
+    # SECURITY - SSL IS OFF
+    - DJANGO_SECRET_KEY=
```
das-g: Isn't this redundant to [our default](https://github.com/geometalab/osmaxx/pull/226/files#diff-0a8ad983bbf48f5cdaa5f2c8c96d0a86R390)?

compose-common.yml
```diff
@@ -12,35 +12,51 @@ base:
...
+    - DJANGO_ALLOWED_HOSTS=
```
das-g: Isn't this redundant to our default?

osmaxx-py/config/settings/common.py
```diff
@@ -388,3 +381,44 @@
...
+CSRF_COOKIE_SECURE = env.bool('DJANGO_CSRF_COOKIE_SECURE', default=False)
+CSRF_COOKIE_HTTPONLY = env.bool('DJANGO_CSRF_COOKIE_HTTPONLY', default=False)
+CSRF_COOKIE_DOMAIN = env.str('DJANGO_CSRF_COOKIE_DOMAIN', default=None)
+CSRF_COOKIE_NAME = env.str('DJANGO_CSRF_COOKIE_NAME', default='csrftoken')
+CSRF_COOKIE_PATH = env.str('DJANGO_CSRF_COOKIE_PATH', default='/')
+CSRF_FAILURE_VIEW = env.str('DJANGO_CSRF_FAILURE_VIEW', default='django.views.csrf.csrf_failure')
+
+## Sessions
+SESSION_CACHE_ALIAS = env.str('DJANGO_SESSION_CACHE_ALIAS', default='default')
+SESSION_COOKIE_AGE = env.int('DJANGO_SESSION_COOKIE_AGE', default=60*60*24*7*2)  # two weeks in seconds
```
das-g: With `from datetime import timedelta`, the default value could be written as

    timedelta(weeks=2).total_seconds()

compose-common.yml
njordan-hsr: You have to set this to a non-empty-string, self documenting, I thought, and I thought wrong. ;-)

compose-common.yml
njordan-hsr: good catch, actually, no, should be 

the settings _above_ only need to be changed...

compose-common.yml
njordan-hsr: And what would that comment need to say? `0.0.0.0 is for listening in any IP`?

I thought that is quite common to write it this way,am I mistaken?

compose-common.yml
njordan-hsr: yes, that way, the environment fetcher (django-environ) sets them to the default when invoked in the settings file.

osmaxx-py/config/settings/common.py
njordan-hsr: Don't know if we really should, it's only used for sending mails to admins and managers (special variants of Djangos `send_mail`). And since we're not using this anywhere, we can just use Django's default which is '[Django] '. But changing is OK too, I don't really care.

osmaxx-py/config/settings/common.py
njordan-hsr: if possible, no imports in settings, other than from other setting modules.

compose-common.yml
njordan-hsr: Yes, it is, but if we don't have this in the `compose-common.yml`, one would have to switch back and forth from the settings-file to know, which env variables are possible at all. I think the better question is, did I maybe forget one here, that can be set?

compose-common.yml
njordan-hsr: Yes, it is, but if we don't have this in the `compose-common.yml`, one would have to switch back and forth from the settings-file to know, which env variables are possible at all. I think the better question is, did I maybe forget one here, that can be set?

osmaxx-py/config/settings/common.py
njordan-hsr: The settings here are the same as Django 1.8 defaults. They are only limited secure. But I see, that this comment is quite misleading.

compose-common.yml
das-g: I guessed it might be something like that. I guess Pythons environment interface is to blame here. It'd be easier to understand if it'd either check for strings representing booleans, like `y`, `yes`, `Off`, etc., or just react on the variable being set or not.

Using this quirk to make the value self-documenting (and documenting the quirk) is a good idea, though. How about

        - PYTHONUNBUFFERED=unbuffered because this string is not empty - see https://docs.python.org/3/using/cmdline.html#envvar-PYTHONUNBUFFERED

compose-common.yml
das-g: It is indeed common in network tools and web frameworks, but back when I joined the project and lacked web-specific development experience, stuff like this was new to me.

compose-common.yml
das-g: If we leave the out in the setting YMLs, they should also be unset, shouldn't they?

osmaxx-py/config/settings/common.py
das-g: Ah, so this doesn't apply to emails we let Django send to end users? (Extractions finished notifications.)

osmaxx-py/config/settings/common.py
das-g: Aw, that's no fun.

But I can see that point. Let's keep it as-is.
:white_check_mark:

setup.production.sh
```diff
@@ -47,7 +69,9 @@ run_docker_compose_dduportal up -d database
 
 sleep 10
 
-# FIXME: create superuser should only be needed to be ran once per deployment setup.
-run_docker_compose_dduportal_interactive run --rm webapp /bin/bash -c "python3 manage.py migrate && python3 manage.py createsuperuser"
+run_docker_compose_dduportal_interactive run --rm webapp /bin/bash -c "./manage.py migrate&&./manage.py collectstatic --noinput"
```
das-g: Please keep spaces around `&&`.

njordan-hsr: Thanks.

e2e/e2e_tests.py
```diff
@@ -1,13 +1,20 @@
...
+from selenium.webdriver.common.by import By
+from selenium.webdriver.support import expected_conditions
+from selenium.webdriver.support.wait import WebDriverWait
 
 try:
     import requests
-    from bs4 import BeautifulSoup  # noqa ignore imported but unused
 except ImportError:
-    print('#### Please install requests and beautifulsoup4, ie. `pip install requests beautifulsoup4` ####')
+    print('#### Please install requests, ie. `pip install requests` ####')
```
das-g: Why treat `requests` differently from `selenium` when both are provided by `e2e/requirements.txt`?

e2e/e2e_tests.py
```diff
@@ -39,105 +46,119 @@ def _stop_containers():
...
-    COOKIES = dict()
-    CSRF_TOKEN = ''
-
-    def _login(self, next=None):
-        login_url = self._make_link('/login/')
-        client = requests.session()
-        client.get(login_url)
+class SeleniumTests(unittest.TestCase):
+    # wait maximal 5 minutes
+    wait_time_seconds = 5*60
```
das-g: While we're at it, let's use `datetime.timedelta(minutes=5).total_seconds()`, here, too. (Unless we want to have compatibility with pre-2.7 python for scripts outside the Docker containers. [`total_seconds` was added in 2.7](https://docs.python.org/2/library/datetime.html#datetime.timedelta).)

e2e/e2e_tests.py
```diff
@@ -39,105 +46,119 @@ def _stop_containers():
...
+
+    def setUp(self):
+        self.browser = self.driver()
+
+    def tearDown(self):
+        self.browser.close()
+
+    def test_login_logout(self):
+        self._login()
+        self._logout()
```
das-g: :+1: for single level of abstraction in this method :smile:

e2e/e2e_tests.py
```diff
@@ -39,105 +46,119 @@ def _stop_containers():
...
-        }
-        r = client.post(login_url, data=login_data)
-        return {'request': r, 'client': client}
+    def test_create_new_excerpt_succeeds(self):
+        self._login()
+        self._create_new_order()
+        self._go_to_order()
+        links = self._wait_for_order_to_complete_and_fetch_download_links()
+        self.assertEqual(len(links), 8)
+        download_links = self._get_download_links_from_links(links)
```
das-g: Method naming here makes not as much sense as it could. As written now it seems like
- You fetch download links and put them into variable `links`
- (you assert that there are 8 of them)
- you get those links from `links` which are download links (which should be all of them, as 'download links' is what you've put in there above) and put them into variable `download_links`.

e2e/e2e_tests.py
```diff
@@ -39,105 +46,119 @@ def _stop_containers():
...
+    def _go_to_order(self):
+        xpath_link_to_order = '/html/body/div/div/div[2]/div/h3/a'
+        WebDriverWait(self.browser, 60).until(
+            expected_conditions.presence_of_element_located((By.XPATH, xpath_link_to_order))
+        )
+        self.browser.find_element_by_xpath(xpath_link_to_order).click()
+
+    def _wait_for_order_to_complete_and_fetch_download_links(self):
+        poll_frequency_in_seconds = 10
+        waited_time = 0
```
das-g: also in seconds, I presume

e2e/e2e_tests.py
```diff
@@ -39,105 +46,119 @@ def _stop_containers():
...
+        status_element = self._get_status_element()
+        while '✓' not in status_element:
+            self.assertFalse('∅' in status_element)
+
+            waited_time += poll_frequency_in_seconds
+            sleep(poll_frequency_in_seconds)
+
+            if waited_time >= self.wait_time_seconds:
```
das-g: You first wait and *then* decide whether you've waited too long? Wouldn't the other way around make more sense?

e2e/e2e_tests.py
```diff
@@ -39,105 +46,119 @@ def _stop_containers():
...
+                raise selenium_exceptions.TimeoutException
+            else:
```
das-g: Unneeded `else` (as the if clause would raise when entered, thereby leaving the loop (and the method) anyway).
Eleminating the `else` (keeping the content of it's clause, of course) would make it easier to see that status_element _will_ be updated as long as we stay in the loop, thus we do have a chance that the loop condition will become false and the loop will terminate.

e2e/e2e_tests.py
```diff
@@ -39,105 +46,119 @@ def _stop_containers():
...
+                self.browser.get(self.browser.current_url)
+                status_element = self._get_status_element()
+        links = self._get_download_a_tags()
+        return links
+
+    def _get_download_links_from_links(self, links):
+        return [links[i].get_attribute('href') for i in range(0, 7, 2)]
```
das-g: Replace magic numbers with constants or expressions based on constants.
##### Suggestion
    NUMBER_OF_LINKS_PER_DOWNLOAD = 2
    EXPECTED_NUMBER_OF_DOWNLAODS = 4
    def _get_download_links_from_links(self, links):
        return [links[i].get_attribute('href') for i in range(0, (NUMBER_OF_LINKS_PER_DOWNLOAD * EXPECTED_NUMBER_OF_DOWNLAODS) - 1, NUMBER_OF_LINKS_PER_DOWNLOAD)]

e2e/e2e_tests.py
```diff
@@ -39,105 +46,118 @@ def _stop_containers():
...
+            expected_conditions.presence_of_element_located((By.ID, "new_excerpt_name"))
+        )
+        for html_id, send_value in self.excerpt_data['send_keys'].items():
+            element = self.browser.find_element_by_id(html_id)
+            element.clear()
+            element.send_keys(send_value)
+        for html_id in self.excerpt_data['click']:
+            self.browser.find_element_by_id(html_id).click()
+        self.browser.find_element_by_xpath("//input[@type='submit']").click()
+        self.browser.get(self._make_link('/orders/'))
```
das-g: `self.browser.get(self._make_link('/orders/'))` belongs into `_go_to_order(self)`

e2e/e2e_tests.py
```diff
@@ -39,105 +46,118 @@ def _stop_containers():
...
-    CSRF_TOKEN = ''
-
-    def _login(self, next=None):
-        login_url = self._make_link('/login/')
-        client = requests.session()
-        client.get(login_url)
+class SeleniumTests(unittest.TestCase):
+    wait_time_seconds = datetime.timedelta(minutes=5).total_seconds()
+    excerpt_data = {
+        'send_keys': {
```
das-g: Can we use a higher-level semantic key name, here, e.g. `form_text_inputs`. That the inputting is done with selenium's `send_keys` is irrelevant for the content of this sub-dict.

e2e/e2e_tests.py
```diff
@@ -39,105 +46,118 @@ def _stop_containers():
...
+            'new_excerpt_name': 'HSR',
+            'new_excerpt_bounding_box_north': '47.22407852727801',
+            'new_excerpt_bounding_box_west':  '8.815616369247437',
+            'new_excerpt_bounding_box_east':  '8.819221258163452',
+            'new_excerpt_bounding_box_south': '47.222388077452706',
+        },
+        'click': [
```
das-g: How about `'checkboxes_to_be_ticked'` as key name here?

e2e/e2e_tests.py
```diff
@@ -39,105 +42,127 @@ def _stop_containers():
...
+
+    def _get_download_links_from_links(self, links):
+        download_links_index_start = 0
+        download_links_index_skip = 2
+        number_of_downloads = 4
+        return [links[i].get_attribute('href') for i in range(
+            start=download_links_index_start,
+            stop=number_of_downloads * download_links_index_skip - 1,
+            step=download_links_index_skip,
+        )]
```
das-g: I noticed we could use slicing here, instead of indexing with a fancy `range`:

    return [link.get_attribute('href') for link in links[download_links_index_start:number_of_downloads * download_links_index_skip - 1:download_links_index_skip]
or even just

    return [link.get_attribute('href') for link in links[download_links_index_start::download_links_index_skip]

njordan-hsr:
Ran tests all tests after last commit (8719bae) with `RUN_E2E=true ./tests.sh` -> All green.

Issue #20, opened 2015-03-10
Webapp: Treat validations errors on service calls as validation errors on the webapp request that triggerd them (create action)


wasabideveloper:
Estimated duration: ~2d

check for

* valid coordinates
* options
* not empty required fields
* ...

in case of constraint violation:
* send user back to form, with his entered data pre-filled (including data violating constraints)
* disable button and indicate constraint violation as in #19 (might implicitly be handled by implementation of #19, anyway)

njordan-hsr:
Maybe a good call for http://django-crispy-forms.readthedocs.org/en/latest/crispy_tag_forms.html#fundamentals ?

das-g:
Validation is done by calling the conversion service (which will itself validate). In this issue, correct reaction to validation errors signalled by the service shall be implemented.

hixi:
@wasabideveloper @das-g 
I think we're not able to integrate this into 1.0 release, is it ok to push this into 2.0?

hixi:
Closing.

PR #234, opened 2015-10-12
Feature/excerpt converter form refactoring

### LINKED ISSUE #236: Fix box not being adapted in Chrome
When selecting an existing order, the box on the map is only changed on Firefox. Make this work on Chrome as well.

njordan-hsr:
Just for review. This is a very cumbersome mission, to be able to simplify the entire thing, one would to refactor some other parts as well - tried to get it just to _match_ the old behavior as close as possible.

Ran test with `RUN_E2E=true ./tests.sh`: all green.

Reviewed by:

* [x] @das-g 
* [ ] @wasabideveloper 

Closes part of #20 
Closes #236 (by accident)


osmaxx-py/osmaxx/excerptexport/forms/excerpt_export_form.py
```diff
@@ -0,0 +1,234 @@
...
+
+    def form_layout(self):
+        return form_layout.Div(
+            form_layout.HTML('<h2>' + _('Existing excerpt') + '</h2>'),
+            form_layout.HTML('''
+                <div class="form-group has-feedback">
+                    <input id="excerptListFilterField" class="form-control" type="search" placeholder="Filter excerpts …" autocomplete="off"/>
+                    <span id="excerptListFilterFieldClearer" class="clearer glyphicon glyphicon-remove-circle form-control-feedback"></span>
+                </div>
+                '''  # noqa: line too long ignored
```
das-g: Should this long string (HTML snipped for the excerpt name filter text input field) be inline in this `form_layout` method or extracted to a class or module constant (or, with the wrapping call to `form_layout.HTML`, to a private method)?

njordan-hsr: I wouldn't worry about this, because it will (read: it has to) disappear in the next round.

das-g: :white_check_mark:

PR #238, opened 2015-10-14
move out of if-statement all that doesn't depend on the condition

das-g:
reviewed by:
- [ ] @njordan-hsr 
- [x] @wasabideveloper


wasabideveloper:
I don't see a dict but it's ok :wink:

das-g:
There was one until before b765223. Removed it due to feedback at https://geometalab.slack.com/archives/osmaxx/p1444825335000026

PR #240, opened 2015-10-14
Feature/add json service for geometries

### LINKED ISSUE #239: Use REST to dynamically fetch data for bounding boxes

njordan-hsr:
REST-API:

* geojson for boundingbox and excerpt (with permissions)
* javascript update to use consume api

Ran tests with `RUN_E2E=true ./test.sh`, all green.

Tested manually:

* back and forth in frontend between exiting and new excerpt

Reviewed by:
- [x] @das-g 
- [x] @wasabideveloper 

Closes #239


osmaxx-py/osmaxx/contrib/auth/frontend_permissions.py
```diff
@@ -45,3 +46,26 @@ class FrontendAccessRequiredMixin(object):
...
+
+    def has_permission(self, request, view):
+        return request.user.is_authenticated() and _may_user_access_osmaxx_frontend(request.user)
+
+
+class HasBBoxAccessPermission(permissions.BasePermission):
+    message = 'Accessing this bounding box is not allowed.'
+
+    def has_object_permission(self, request, view, obj):
+        return obj.excerpt.is_public or obj.excerpt.owner == request.user
```
das-g: This is essentially

    HasExcerptAccessPermission#has_object_permission(
        request=request,
        view=view,
        obj=obj.excerpt,
    )

While actually calling that would be more code (would we have to instantiate `HasExcerptAccessPermission`?) keeping that logic in only one place might still be benefitial if the duplication isn't coincidental. (And I don't think it is.)

osmaxx-py/osmaxx/excerptexport/models/bounding_geometry.py
```diff
@@ -25,6 +32,15 @@ class OsmosisPolygonFilterBoundingGeometry(BoundingGeometry):
...
     polygon_file = models.FileField(storage=get_private_upload_storage())
 
+    def geometry(self):
+        # fake polygon for now
+        return Polygon([
+            [0, 1],
+            [1, 0],
+            [1, 1],
+            [0, 1],
+        ])
```
das-g: Should we still return a fake here?

osmaxx-py/osmaxx/excerptexport/static/excerptexport/libraries/Leaflet.Spin/leaflet.spin.js
```diff
@@ -0,0 +1,41 @@
+L.SpinMapMixin = {
+    spin: function (state, options) {
+        if (!!state) {
```
das-g: `!!` should not be required here, `if` implicitly casts to Boolean, doesn't it?

osmaxx-py/osmaxx/excerptexport/static/excerptexport/libraries/Leaflet.Spin/leaflet.spin.js
```diff
@@ -0,0 +1,41 @@
+L.SpinMapMixin = {
+    spin: function (state, options) {
```
das-g: Avoid flag arguments (smell/heuristic **F4**, in the 'Clean Code' book) / selector arguments (smell/heuristic **G15**). As you never call this with a state that isn't a boolean literal, split this function into two: `start_spinning` and `stop_spinning`.

osmaxx-py/osmaxx/excerptexport/rest_api/serializers.py
```diff
@@ -0,0 +1,41 @@
...
+from osmaxx.excerptexport.models import Excerpt
+from osmaxx.excerptexport.models.bounding_geometry import BoundingGeometry
+from osmaxx.utilities.shortcuts import get_actual
+
+
+class BoundingGeometrySerializer(gis_serializers.GeoFeatureModelSerializer):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    geom = GeometrySerializerMethodField()
```
wasabideveloper: I would prefer long names for variables.

osmaxx-py/osmaxx/excerptexport/static/excerptexport/libraries/Leaflet.Spin/leaflet.spin.js
```diff
@@ -0,0 +1,41 @@
...
+        if (!!state) {
+            // start spinning !
+            if (!this._spinner) {
+                this._spinner = new Spinner(options).spin(this._container);
+                this._spinning = 0;
+            }
+            this._spinning++;
```
das-g: Can there be multiple reasons to spin at a given moment and you have to keep spinning until all of them are gone?

osmaxx-py/osmaxx/utilities/shortcuts.py
```diff
@@ -32,3 +32,16 @@ def get_list_or_none(klass, *args, **kwargs):
     if not object_list:
         return None
     return object_list
+
+
+def get_actual(obj):
+    """Expands `obj` to the actual object type.
+    """
```
wasabideveloper: Is it necessary to break the closing """ on a new line?

osmaxx-py/osmaxx/excerptexport/static/excerptexport/libraries/Leaflet.Spin/leaflet.spin.js
```diff
@@ -0,0 +1,41 @@
...
+        // If added layer is currently loading, spin !
+        if (e.layer.loading) this.spin(true);
+        if (typeof e.layer.on != 'function') return;
+        e.layer.on('data:loading', function () { this.spin(true); }, this);
+        e.layer.on('data:loaded',  function () { this.spin(false); }, this);
+    }, this);
+    this.on('layerremove', function (e) {
+        // Clean-up
+        if (e.layer.loading) this.spin(false);
+        if (typeof e.layer.on != 'function') return;
```
das-g: Always use curly braces with `if`, even when only one statement follows.

And why do we care whether `on` is a function for calling `off`? If anything, wouldn't we have to make sure that `off` is a function?

osmaxx-py/osmaxx/excerptexport/static/excerptexport/libraries/Leaflet.Spin/leaflet.spin.js
```diff
@@ -0,0 +1,41 @@
...
+    }
+};
+
+L.Map.include(L.SpinMapMixin);
+
+L.Map.addInitHook(function () {
+    this.on('layeradd', function (e) {
+        // If added layer is currently loading, spin !
+        if (e.layer.loading) this.spin(true);
+        if (typeof e.layer.on != 'function') return;
```
das-g: When would `on` not be a function?

osmaxx-py/osmaxx/excerptexport/static/excerptexport/scripts/map.js
```diff
@@ -72,25 +65,29 @@
...
+            this._setLocationFilterFromExcerptID(excerptOption.value);
         }.bind(this));
 
         // enable excerpt on map on change of form mode (existing excerpt or new excerpt)
         this.formElementPartsSwitcher.addEventListener('change', function(event) {
             this.locationFilter.enable();
         }.bind(this));
+
+        this._setLocationFilterFromExcerptID = function(ID) {
+            var that = this;
```
das-g: Isn't it conventional to use a variable named `self` as alias for `this` for closure purposes?

osmaxx-py/osmaxx/utilities/shortcuts.py
```diff
@@ -32,3 +32,16 @@ def get_list_or_none(klass, *args, **kwargs):
     if not object_list:
         return None
     return object_list
+
+
+def get_actual(obj):
+    """Expands `obj` to the actual object type.
```
das-g: Documentation of this function should probably work with inheritance-related terms like 'parent type' / 'child type' or 'super type' / 'subtype', or if this works for indirect inheritance 'ancestor type' / 'descendent type'.

osmaxx-py/osmaxx/utilities/shortcuts.py
```diff
@@ -32,3 +32,16 @@ def get_list_or_none(klass, *args, **kwargs):
...
+    """
+    for name in dir(obj):
+        try:
+            attr = getattr(obj, name)
+            if isinstance(attr, obj.__class__):
```
das-g: In general, this condition isn't sufficient an may lead to the wrong attribute being returned.

das-g:
Should any of the new code be unit- or module-tested? Are new integration- or end-to-end tests necessary?

osmaxx-py/osmaxx/excerptexport/static/excerptexport/libraries/Leaflet.Spin/leaflet.spin.js
njordan-hsr: You might point that out upstream: https://github.com/makinacorpus/Leaflet.Spin

osmaxx-py/osmaxx/excerptexport/static/excerptexport/libraries/Leaflet.Spin/leaflet.spin.js
njordan-hsr: You might point that out upstream: https://github.com/makinacorpus/Leaflet.Spin

osmaxx-py/osmaxx/excerptexport/static/excerptexport/libraries/Leaflet.Spin/leaflet.spin.js
njordan-hsr: You might point that out upstream: https://github.com/makinacorpus/Leaflet.Spin

osmaxx-py/osmaxx/utilities/shortcuts.py
njordan-hsr: No, but the comment should actually be expanded to more than one line explaining what this function does.

osmaxx-py/osmaxx/excerptexport/static/excerptexport/libraries/Leaflet.Spin/leaflet.spin.js
njordan-hsr: You might point that out upstream: https://github.com/makinacorpus/Leaflet.Spin

osmaxx-py/osmaxx/excerptexport/static/excerptexport/libraries/Leaflet.Spin/leaflet.spin.js
njordan-hsr: You might point that out upstream: https://github.com/makinacorpus/Leaflet.Spin

njordan-hsr:
Please wait with reviewing of ed844fa to 94ba2e2 -  I broke the test suite and need to fix that first. Along fixing that, I'll probably have to ad some new tests for the additional code as well.

osmaxx-py/osmaxx/contrib/auth/frontend_permissions.py
njordan-hsr: yes, we would need to instantiate it, but we can always just extract the function, e.g
```python
def has_user_access_to_excerpt(user, excerpt):
    return excerpt.is_public or excerpt.owner == user
```

osmaxx-py/osmaxx/excerptexport/rest_api/serializers.py
njordan-hsr: could rename it geometry now, since that name is no longer being used by any other class. Other propositions?

osmaxx-py/osmaxx/utilities/shortcuts.py
njordan-hsr: It has all sorts of problems (that one just being one of them), if you know of a cleaner solution, I'm all ears.

osmaxx-py/osmaxx/utilities/shortcuts.py
njordan-hsr: This function should probably not only be better documented, but also point out it's weaknesses: it can easily lead to stack overflows (by declaring a property that uses the `get_actual` as reference) and the issue you pointed out below.

osmaxx-py/osmaxx/excerptexport/static/excerptexport/libraries/Leaflet.Spin/leaflet.spin.js
das-g: Oops, I didn't notice that this was in `osmaxx/osmaxx-py/osmaxx/excerptexport/static/excerptexport/`**`libraries`** and thought this was a script written by us.

osmaxx-py/osmaxx/utilities/shortcuts.py
das-g: Can https://django-model-utils.readthedocs.org/en/latest/managers.html#inheritancemanager be of help, here?

osmaxx-py/osmaxx/contrib/auth/frontend_permissions.py
das-g: Good idea. Would it make sense to have this method in the excerpt model? (Or do we have more models that folow the `.is_public`-and-`.owner` ducktype interface? Then it might make sense to have the method in a common super class.)

osmaxx-py/osmaxx/excerptexport/rest_api/serializers.py
das-g: I'm fine with both renaming to `geometry`.

osmaxx-py/osmaxx/contrib/auth/frontend_permissions.py
njordan-hsr: Yes, that would make sense. It would actually make sense to not have names like `orderer`, but actually always use `owner` or alternatively `user`, though I prefer the former. An abstract base class would most likely make sense here, which has both: an owner field and the _duck-interface_.

osmaxx-py/osmaxx/utilities/shortcuts.py
njordan-hsr: I like that one, much cleaner:

```python
        for rel in rels:
            if levels or levels is None:
                for subclass in self._get_subclasses_recurse(
                        rel.field.model, levels=levels):
                    subclasses.append(
                        rel.get_accessor_name() + LOOKUP_SEP + subclass)
            subclasses.append(rel.get_accessor_name())
```

osmaxx-py/osmaxx/contrib/auth/frontend_permissions.py
njordan-hsr: On other thoughts, permissions should not be a concern of the model itself, since it can be viewed as orthogonal to the model itself. I'm therefore placing it in a method in permissions, agree?

das-g: I think I'd prefer to have the actual logic on the model (be it on the specific class or on a super class) and have a relatively 'dumb' method calling it (similar complexity to `_may_user_access_osmaxx_frontend`) in `frontend_permissions.py`. (Not placing the actual logic in `frontend_permissions.py` avoids 'Clean Code' smell G14. Somewhat similar to "Tell, don't ask.")

As our models are not intended to be used in other applications (that might have different access regimes), I don't think we'd obstruct anything by having this logic in the model.

njordan-hsr: There also 'Clean Code' G17, as a contrast, and one would expect permissions to be in the permissions module, not?

osmaxx-py/osmaxx/excerptexport/models/bounding_geometry.py
njordan-hsr: yes, the rest comes with another branch, implementing this feature.

osmaxx-py/osmaxx/utilities/shortcuts.py
njordan-hsr: Will be removed.

osmaxx-py/osmaxx/excerptexport/static/excerptexport/scripts/map.js
njordan-hsr: I've seen both out in the wild, don't know which one is more common. I think I borrowed this `pattern` from a codesnippet example somewhere, as inspiration. [This guy](http://stackoverflow.com/questions/2698831/that-self-or-me-which-one-to-prefer-in-javascript) uses `that` personally in the accepted answer, and makes an example with self.

njordan-hsr:
Ran `RUN_E2E=true ./test.py` after last commit (5eba305) and all green.

Manually tested basic extraction of excerpts, seems ok.

No new tests added, since the only real new code is configuration of ClassBasedViews for the API. Maybe at a later stage, when we actually use the api (more), we should test the REST API on higher level (ie. is it returning the right _stuff_). This makes sense, at the point where we are happy with the interface it has.

osmaxx-py/osmaxx/excerptexport/static/excerptexport/scripts/map.js
das-g: :white_check_mark:

njordan-hsr:
Ran `RUN_E2E=true ./test.py` after last commit (164f28e) and all green.

Manually tested basic extraction of excerpts, seems ok.

No new tests added, since the only real new code is configuration of ClassBasedViews for the API. Maybe at a later stage, when we actually use the api (more), we should test the REST API on higher level (ie. is it returning the right _stuff_). This makes sense, at the point where we are happy with the interface it has.

osmaxx-py/osmaxx/excerptexport/rest_api/serializers.py
```diff
@@ -0,0 +1,38 @@
...
+
+    proxy_for_geometry = GeometrySerializerMethodField()
+
+    def get_proxy_for_geometry(self, obj):
+        return obj.subclass_instance.geometry
+
+    class Meta:
+        geo_field = 'proxy_for_geometry'
+        model = BoundingGeometry
+        fields = ['id', 'type_of_geometry']
```
das-g: What is `type_of_geometry` used for by the client? Do we need it at all?

osmaxx-py/osmaxx/excerptexport/rest_api/serializers.py
```diff
@@ -0,0 +1,38 @@
...
+    proxy_for_geometry = GeometrySerializerMethodField()
+
+    def get_proxy_for_geometry(self, obj):
+        bounding_geometry = obj.bounding_geometry
+        return bounding_geometry.subclass_instance.geometry
+
+    class Meta:
+        geo_field = 'proxy_for_geometry'
+        model = Excerpt
+        fields = ['id', 'type_of_geometry']
```
das-g: What is `type_of_geometry` used for by the client? Do we need it at all?

osmaxx-py/osmaxx/excerptexport/static/excerptexport/scripts/map.js
```diff
@@ -31,16 +31,10 @@
         };
 
         this.isSelectOptionSelectedAndExcerptOnMapInSyncWithInputFields = function(select, locationFilterBounds) {
-            if(select.value == "") {
+            if (select.value == "" || this.selectedExcerptGeoJson == null) {
```
das-g: Is it still required to check `select.value` if it isn't used anymore in the remainder of this function?

osmaxx-py/osmaxx/excerptexport/models/bounding_geometry.py
```diff
@@ -10,13 +13,23 @@ def type(self):
...
     @property
-    def geometry_instance(self):
-        if hasattr(self, 'bboxboundinggeometry'):
-            return self.bboxboundinggeometry
-        elif hasattr(self, 'osmosispolygonfilterboundinggeometry'):
-            return self.osmosispolygonfilterboundinggeometry
-        else:
-            return self
+    def geometry(self):
+        raise NotImplementedError
```
das-g: Would it make sense to return `self.subclass_instance.geometry` here? This would simplify the implementation of `type_of_geometry`.

njordan-hsr: yes, it would.

osmaxx-py/osmaxx/excerptexport/static/excerptexport/scripts/map.js
njordan-hsr: done

njordan-hsr:
Ran `RUN_E2E=true ./test.py` after last commit (f64e642): all green.

PR #242, opened 2015-10-16
get rid of git submodules

das-g:
(cleans up after #227)

#### reviewed by:
- [ ] @wasabideveloper
- [ ] @njordan-hsr

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,11 +100,6 @@ See Wiki: https://github.com/geometalab/osmaxx/wiki
 
 ### Problems & Solutions
 
-#### Leaflet is missing
-
-You forgot to **pull the submodules**
-
-
 #### ProgrammingError at /login/
 
 ```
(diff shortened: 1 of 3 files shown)


wasabideveloper:
What's the reason for checking in foreign code in our repo and removing the submodule?

das-g:
> What's the reason for checking in foreign code in our repo and removing the submodule?

I guess:

* Simplifying setup for new developers / new clones of the repo
* Simplifying deployment

@njordan-hsr can you comment on that? You've made the changes in #227

wasabideveloper:
I think it's not a good idea to remove the submodules. It's the way, subrepos/foreign code should be handled. For the deployment we use a script and for development setup we can do the same. So it's not necessary to simplify this a little bit by doing something bad on another side.

njordan-hsr:
We used a master checkout from a github repo, which might change at any point in the future. We could've locked it to a certain commit, but in general we don't use it much differently as any other third party javascript code, so I think it makes sense to have all using the same mechanism (checking it into source control and be sure, it is available in the version we want).

PR #271, opened 2015-11-25
Don't require runtime-related ENV variables for common development tasks

### LINKED ISSUE #270: Don't require runtime-related ENV variables for common development tasks
As a developer, I want to be able to perform common development tasks, such as creating migrations, without having to add my source into a docker container as a volume (in fact without having to use docker at all). Doing something like

    virtualenv-3.4 .venv
    . .venv/bin/activate
    pip3 install -r osmaxx-py/requirements/local.txt
    osmaxx-py/manage.py makemigrations
currently fails with `django.core.exceptions.ImproperlyConfigured: Set the DJANGO_OSMAXX_CONVERSION_SERVICE_USERNAME environment variable`

However, for running migrations this variable is completely irrelevant, so setting should not be required.

#### Note:
Fixing this shouldn't change that forgetting to set this variable in **production** causes an early failure to draw attention to that omission.

das-g:
Don't require runtime-related ENV variables for common development tasks like like creating migrations

fixes #270


hixi:
Like this, ConversionApiClient which gets `CONVERSION_SERVICE_USERNAME` and `CONVERSION_SERVICE_PASSWORD` will fail locally or on testing, don't know if that is what we want.

osmaxx-py/config/settings/common.py
```diff
@@ -381,8 +381,6 @@
     # to request access to osmaxx (access_denied page)
     'account_manager_username': env.str('ACCOUNT_MANAGER_USERNAME', default='admin'),
     'CONVERSION_SERVICE_URL': env.str('DJANGO_OSMAXX_CONVERSION_SERVICE_URL', default='http://localhost:8901/api/'),
-    'CONVERSION_SERVICE_USERNAME': env.str('DJANGO_OSMAXX_CONVERSION_SERVICE_USERNAME'),
```
hixi: would `env.str('DJANGO_OSMAXX_CONVERSION_SERVICE_USERNAME', default='unknown_user)` be acceptable?

osmaxx-py/config/settings/common.py
```diff
@@ -381,8 +381,6 @@
...
-    'CONVERSION_SERVICE_PASSWORD': env.str('DJANGO_OSMAXX_CONVERSION_SERVICE_PASSWORD'),
```
hixi: would `env.str('DJANGO_OSMAXX_CONVERSION_SERVICE_PASSWORD', default='unknown_password)` be acceptable? (or something similar)

das-g: Acceptable, yes, because even if this password would happen stay unreplaced on production, it would just cause the UI to try to use it to access the service. It would not cause the service or UI to be accessible using this default password.

But still a bit suboptimal from a debugging perspective.

das-g:
> Like this, ConversionApiClient which gets `CONVERSION_SERVICE_USERNAME` and `CONVERSION_SERVICE_PASSWORD` will fail locally or on testing, don't know if that is what we want.

Testing shouldn't be affected, as `config/settings/test.py` also sets these options (hardcoded, without regard to any ENV variables). But good point about local running.

osmaxx-py/config/settings/common.py
hixi: on production, it will bug us, because of your change in `production.py`

hixi:
You're right about testing.py, but for running locally I think it'd be a good idea non-the-less.

osmaxx-py/config/settings/common.py
das-g: Ah, right, `production` overrides `common`. Will change it as you suggested.

das-g: done: 1620a8d

PR #272, opened 2015-11-27
Feature/fix e2e tests

hixi:
Fixes the e2e test not running; fixes testing issues with normal testing as well.

Reviewed by:
- [x] @das-g 

Don't review the yml files, they are autogenerated.


compose-common.yml
```diff
@@ -29,3 +29,33 @@ base:
     # We enable Python hash randomization to protect Django from certain DoS attacks.
     # See https://docs.djangoproject.com/en/1.8/howto/deployment/checklist/#python-options
     - PYTHONHASHSEED=random
+
```
das-g: If above `# These settings probably shouldn't need a change` is valid for the whole rest of the file, it should be un-indented (if YAML allows that)

hixi: :white_check_mark:

compose-common.yml
```diff
@@ -29,3 +29,39 @@ base:
...
+    # hardened security compared to default settings.
+    - DJANGO_SECURE_BROWSER_XSS_FILTER=true
+    - DJANGO_SECURE_CONTENT_TYPE_NOSNIFF=true
+    - DJANGO_LOG_LEVEL=ERROR
+
+    # Logging
+    - PYTHONUNBUFFERED=non-empty-string
+    - PYTHONIOENCODING=utf-8
+    - LC_ALL=C.UTF-8
+    - LANG=C.UTF-8
```
das-g: Do this locale/language settings belong with logging or would they belong in a blank-line-separated block (with or without comment) of their own?

osmaxx-py/osmaxx/excerptexport/tests/test_conversion_api_client.py
```diff
@@ -79,27 +80,31 @@ def test_create_job(self):
...
-            len(self.extraction_order.output_files.order_by('id')[1].file.read()),
-            368378
-        )
+        with vcr.use_cassette('fixtures/vcr/conversion_api-test_download_files.yml') as cassette:
+            api_client = ConversionApiClient()
+            api_client.create_job(self.extraction_order)
+
+            if cassette.dirty:
+                # wait for external service to complete request
+                time.sleep(120)
```
das-g: Looking at [the code of `Cassette`](https://github.com/kevin1024/vcrpy/blob/1324feae9920e449e44a987ad178a8ec799a413f/vcr/cassette.py#L148), our use of `dirty` (indicating whether the last response came from the record or has been freshly recorded just now) doesn't match that attribute's intended purpose (indicating whether the cassette object contains unsaved request-response pairs, i.e., not yet persisted to the cassette file).

If we want to keep (ab)using `dirty`, we should have a separate test in place that ensures it keeps working as we want.

das-g: Alternatives might be to use `play_count` or to simply check whether the cassette file is present.

osmaxx-py/osmaxx/excerptexport/tests/test_conversion_api_client.py
```diff
@@ -79,27 +80,31 @@ def test_create_job(self):
...
+
+            success = api_client.download_result_files(self.extraction_order)
+            self.assertIsNone(api_client.errors)
+            self.assertTrue(success)
+            self.assertEqual(self.extraction_order.output_files.count(), 2)
+            self.assertEqual(self.extraction_order.output_files.order_by('id')[0].content_type, 'fgdb')
+            self.assertEqual(self.extraction_order.output_files.order_by('id')[1].content_type, 'spatialite')
```
das-g: Unchanged in this PR, but do we care about whether the formats get their IDs assigned in order?

If not, the above 3 asserts could be boiled down to something like

    self.assertCountEqual(
        outfile.content_type for outfile in self.extraction_order.output_files,
        ('fgdb', 'spatialite')
    )

runtests.py
```diff
@@ -130,9 +114,20 @@ def application_tests(self):
         self.log_header('Application tests:')
 
         try:
-            self.log_docker_compose(['run', self.WEBAPP_CONTAINER, '/bin/bash', '-c',
-                                     'DJANGO_SETTINGS_MODULE=config.settings.test python3 manage.py test'])
-            self.log_success("Tests passed successfully")
+            import django  # noqa
+        except ImportError:
+            print('Are you in a activated virtualenv and have installed the settings?')
```
das-g: "installed the _settings_" ? Did you mean "requirements"?

compose-common.yml
hixi: They have to do with the logging - outputting with `PYTHONIOENCODING=utf-8` and not `LC_ALL=C.UTF-8` `LANG=C.UTF-8` seometimes results in quite strange error outputs from the docker container.

osmaxx-py/osmaxx/excerptexport/tests/test_conversion_api_client.py
hixi: I think it would make sense. Can you open an issue for this?

das-g: :white_check_mark: Issue opened: #273

compose-common.yml
das-g: :white_check_mark: leave as-is then

Issue #257, opened 2015-11-12
Worker: send url for status query as payload


hixi:
Refactor the worker, so it can send the url for the status update with the get request using the callback url.
(send url for status query as payload)

Issue #247, opened 2015-10-19
Webapp: React on status change callback


ghost:
on status update (i.e. when callback received)
- [x] use url from payload to request current status for the respective format (depends on #257)
- [x] update own status in database
    - [x] if changed: notify user (within web UI)
-  when status == finished
    - [x] retrieve file for respective format (use URL from status request answer) (retry if necessary?)
    - [ ] ~~request file deletion on other side if retrieval was successful (use URL from status request answer)~~ (tracked by #279 , #139 and #289 )
- when all formats of order have a final status (finished or error)
    - [x] notify user (by email)

das-g:
Depends on depends on #257

PR #280, opened 2015-12-02
Feature/247 react on status change callback

das-g:
First part of #247:
- [x] use url from payload to request current status for the respective format (depends on #257)
- [x] update own status in database

Also refactored some existing tests.

Reviewed by:
- [x] @hixi
- [x] @wasabideveloper


osmaxx-py/osmaxx/excerptexport/services/conversion_api_client.py
```diff
@@ -53,7 +54,9 @@ def create_job(self, extraction_order):
         bounding_geometry = extraction_order.excerpt.bounding_geometry.subclass_instance
 
         request_data = OrderedDict({
-            "callback_url": "http://example.com",
+            "callback_url": "http://example.com{0}".format(
```
hixi: we should ensure that this is in fact not _http://example.com_ in production. Would this change belong in this pull request as well?

osmaxx-py/osmaxx/job_progress/urls.py
```diff
@@ -0,0 +1,7 @@
+from django.conf.urls import url
+
+from osmaxx.job_progress import views
+
+urlpatterns = [
+    url(r'^tracker/(?P<order_id>[0-9]+)$', views.tracker, name='tracker'),
```
hixi: it is quite usual to have a trailing slash in the django urls ([though this can be changed](https://docs.djangoproject.com/en/dev/ref/settings/#append-slash)). 
I think we should stick with the default here.

das-g: fixed with d4db26a

osmaxx-py/osmaxx/excerptexport/services/conversion_api_client.py
das-g: Yes, I'd say so. PRs don't have to be feature complete, but they should be production ready.

wasabideveloper: Do we not already store the domain in the configuration? Can we use that entry?

hixi: No, we don't (at least _not really_ :tm: ). In the case of running within docker in production, we _could_ take it from the `VIRTUAL_HOST` docker environment variable, but that is very specific how we run this image(s).

Two options: 
- defining the a site in Django's site framework, and take it from there. This would also be trivial to extend the entrypoint script for the docker-container, and setting that `VIRTUAL_HOST` (or any other environment entry that suits us) to the corresponding [site](https://docs.djangoproject.com/en/1.8/ref/contrib/sites/) and [SITES_ID](https://docs.djangoproject.com/en/1.8/ref/settings/#std:setting-SITE_ID) to 1.
- somehow reaching/getting to the request and then using django-rest-frameworks [reverse](http://www.django-rest-framework.org/api-guide/reverse/#reverse), which returns the fully qualified url.

hixi: After looking at the code again, I'm in favor of the first option.

wasabideveloper: :+1:

das-g: Implemented second option (host from request) with bc4d8c8.

PR #282, opened 2015-12-04
Fix not working select of existing excerpt

### LINKED ISSUE #185: Fix not working select on select existing excerpt the second time
## To reproduce:
1. Select an existing excerpt from the list by clicking its name in the list
2. Modify the excerpt on the map &rarr; the application will change to the 'new excerpt' form
3. Change back to the list of existing excerpt
4. Try to re-choose the previously chosen excerpt by clicking on its name in the list

### expected:
* The excerpt on the map snaps back to the extend of the chosen excerpt.

### observed:
* Nothing happens.

Estimated duration: ~1d

wasabideveloper:
Implementation of fix #185 (Merge will close #185 ).

* Tested in IE11, FF and Chrome and it works well

To be reviewed by:
- [x] @das-g 
- [x] @hixi

### DIFF
--- a/osmaxx-py/osmaxx/excerptexport/static/excerptexport/scripts/map.js
+++ b/osmaxx-py/osmaxx/excerptexport/static/excerptexport/scripts/map.js
@@ -84,7 +84,7 @@
 
         this._setLocationFilterFromExcerptID = function(ID) {
             var that = this;
-            this.selectedExcerptGeoJson = L.geoJson.ajax("/api/bounding_geometry_from_excerpt/"+ID+"/?format=json").on('data:loaded', function(){
+            this.selectedExcerptGeoJson = L.geoJson.ajax("/api/bounding_geometry_from_excerpt/"+ID+"/").on('data:loaded', function(){
         //TODO: differentiate between boundingbox or country and similar; use this.locationFilter.enable()/disable()
                 that.locationFilter.setBounds(this.getBounds());
                 map.fitBounds(that.locationFilter.getBounds());
--- a/osmaxx-py/osmaxx/excerptexport/static/excerptexport/scripts/map.js
+++ b/osmaxx-py/osmaxx/excerptexport/static/excerptexport/scripts/map.js
@@ -60,12 +60,22 @@
             this.inputElementsNewBoundingBox[inputElementKey].addEventListener('change', this.updateMapExcerpt.bind(this));
         }.bind(this));
 
+
         // update excerpt on map on selection of existing excerpt in list
-        this.selectElementExistingExcerpts.addEventListener('change', function() {
+        var getExcerptOptionAndUpdateFilter = function() {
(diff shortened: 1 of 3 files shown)


osmaxx-py/osmaxx/excerptexport/static/excerptexport/scripts/map.js
```diff
@@ -74,7 +84,7 @@
 
         this._setLocationFilterFromExcerptID = function(ID) {
             var that = this;
-            this.selectedExcerptGeoJson = L.geoJson.ajax("/api/bounding_geometry_from_excerpt/"+ID+"/").on('data:loaded', function(){
+            this.selectedExcerptGeoJson = L.geoJson.ajax("/api/bounding_geometry_from_excerpt/"+ID+"/?format=json").on('data:loaded', function(){
```
das-g: @hixi Should clients need to specify `format=json`? I'd expect this to be the default if the client doesn't specify any preferred content types. (Or _does_ `L.geoJson.ajax` specify some HTML content types as preferred, thus making it necessary to override that with an URL parameter?)

das-g:
Has this been tested on multiple browsers? (I have no idea how portable or browser-specific the used JavaScript features are.)

osmaxx-py/osmaxx/excerptexport/static/excerptexport/scripts/map.js
hixi: No, it shouldn't. If it is indeed needed right now, I'd consider this a bug on the osamaxx rest api client. 

We probably just haven't set the settings:

```python
REST_FRAMEWORK = {
    'DEFAULT_RENDERER_CLASSES': (
        'rest_framework.renderers.JSONRenderer',
    ),
    'DEFAULT_PARSER_CLASSES': (
        'rest_framework.parsers.JSONParser',
    )
}
```

See: [REST Framework Settings](http://www.django-rest-framework.org/api-guide/settings/#settings)

wasabideveloper: It was not working without specifying `format=json` so I hat to use that. I'm open for a other solution...

hixi: See comment above ;-)

wasabideveloper: :heavy_check_mark:

wasabideveloper:
I'm testing on IE. But there we have some strange behaviour anyway. I will try to improve it.

wasabideveloper:
@das-g and @hixi please take a look again. I replaced `formPartsSwitcher.dispatchEvent(new Event('change'));` by my own method (&rarr;extract callback to onFormPartsSwitcherChange) to support IE and enhance decoupling.

hixi:
:+1:

Issue #154, opened 2015-07-29
Converter: Implement Garmin excerpt conversion


wasabideveloper:
**Estimated duration**: ~4d

- [ ] Django/Celery code
- [ ] &nbsp;(if necessary) separate docker container

![Data flow chart](https://raw.githubusercontent.com/geometalab/osmaxx-wiki-vault/master/data%20flows%20and%20tools.png)

Copy of the "old" export options including Garmin:
```python
EXPORT_OPTIONS = {
    'gis': {
        'name': 'GIS',
        'formats': {
            'file_gdb': {
                'name': 'FileGDB',
                'file_extension': 'gdb',
                'mime_type': 'application/octet-stream'
            },
            'geo_package': {
                'name': 'GeoPackage',
                'file_extension': 'gpkg',
                'mime_type': 'application/octet-stream'  # Not verified!
            },
            'shape_file': {
                'name': 'ShapeFile',
                'file_extension': 'shp',
                'mime_type': 'application/octet-stream'  # Not verified!
            }
        },
        'options': {
            'coordinate_reference_system': {
                'label': 'Coordinate reference system',
                'type': 'choice',
                'default': 'pseudomerkator',
                'groups': [
                    {
                        'name': 'Global coordinate reference systems',
                        'values': [
                            {'name': 'pseudomerkator', 'label': 'Pseudo merkator'},
                            {'name': 'wgs72', 'label': 'WGS 72'},
                            {'name': 'wgs84', 'label': 'WGS 84'}
                        ]
                    },
                    {
                        'name': 'UTM zones for your export',
                        'values': [
                            {'name': 'utm32', 'label': 'UTM zone 32'},
                            {'name': 'utm33', 'label': 'UTM zone 33'}
                        ]
                    }
                ]
            },
            'detail_level': {
                'label': 'Detail level',
                'type': 'choice',
                'default': 'verbatim',
                'values': [
                    {'name': 'verbatim', 'label': 'Verbatim'},
                    {'name': 'simplified', 'label': 'Simplified'}
                ]
            }
        }
    },
    'routing': {
        'name': 'Routing',
        'formats': {
            'img': {
                'name': 'IMG (Garmin)',
                'file_extension': 'img',
                'mime_type': 'application/octet-stream'  # Not verified!
            }
        },
        'options': {

        }
    }
}
```

hixi:
This issue must be reworked, since a lot changed and moved to the conversion-service.

PR #292, opened 2015-12-08
Feature/247 react on status change callback

### LINKED ISSUE #247: Webapp: React on status change callback

hixi:
Reaction on status callback and user messaging in case of change.

Reviewed by:

- [x] @wasabideveloper 
- [x] @das-g 

closes #247

### DIFF
--- a/docs/project-development-environment.md
+++ b/docs/project-development-environment.md
@@ -55,7 +55,9 @@ To run the application tests only, see [Commonly used commands while developing 
 
 ## Access the application
 
-[http://localhost:8000](http://localhost:8000)
+`http://<your_ip>:8000`
+
+where `<your_ip>` is your (public) IP.
 
 or add
 
--- a/docs/project-development-environment.md
+++ b/docs/project-development-environment.md
@@ -57,7 +57,10 @@ To run the application tests only, see [Commonly used commands while developing 
 
 `http://<your_ip>:8000`
 
-where `<your_ip>` is your (public) IP.
+where `<your_ip>` is your (public) IP as reported by
+```bash
+ip route get 1 | awk '{print $NF;exit}'
+```
 
 or add
 
--- a/osmaxx-py/osmaxx/excerptexport/migrations/0012_outputfile_file_status.py
+++ b/osmaxx-py/osmaxx/excerptexport/migrations/0012_outputfile_file_status.py
@@ -1,0 +1,19 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('excerptexport', '0011_remove_extractionorder_process_start_date'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='outputfile',
+            name='file_status',
+            field=models.IntegerField(verbose_name='file status', choices=[(1, 'downloading'), (2, 'received')], default=0),
+        ),
+    ]
--- a/osmaxx-py/osmaxx/excerptexport/models/extraction_order.py
+++ b/osmaxx-py/osmaxx/excerptexport/models/extraction_order.py
@@ -19,6 +19,19 @@ class ExtractionOrderState(enum.Enum):
     FAILED = 6
 
 
+CONVERSION_PROGRESS_TO_EXTRACTION_ORDER_STATE_MAPPING = {
+    'new': ExtractionOrderState.INITIALIZED,
+    'received': ExtractionOrderState.QUEUED,
+    'started': ExtractionOrderState.PROCESSING,
+    'successful': ExtractionOrderState.FINISHED,
+    'error': ExtractionOrderState.FAILED,
+}
+
+
+def get_order_status_from_conversion_progress(progress):
+    return CONVERSION_PROGRESS_TO_EXTRACTION_ORDER_STATE_MAPPING.get(progress, ExtractionOrderState.UNDEFINED)
+
+
 class ExtractionOrder(models.Model):
     state = enum.EnumField(ExtractionOrderState, default=ExtractionOrderState.INITIALIZED, verbose_name=_('state'))
     _extraction_configuration = models.TextField(
@@ -64,3 +77,9 @@ class ExtractionOrder(models.Model):
     @property
     def extraction_formats(self):
         return json.loads(self.extraction_configuration).get('gis_formats', None)
+
+    def set_status_from_conversion_progress(self, job_overall_progress):
+        # don't update if we already we're done or errored out
+        if self.state not in [ExtractionOrderState.FINISHED, ExtractionOrderState.FAILED]:
+            self.state = get_order_status_from_conversion_progress(job_overall_progress)
+            self.save()
--- a/osmaxx-py/osmaxx/excerptexport/models/extraction_order.py
+++ b/osmaxx-py/osmaxx/excerptexport/models/extraction_order.py
@@ -90,6 +90,5 @@ class ExtractionOrder(models.Model):
         return json.loads(self.extraction_configuration).get('gis_formats', None)
 
     def set_status_from_conversion_progress(self, job_overall_progress):
-        # Don't update if we're already done or have errored out:
         if self.state not in [ExtractionOrderState.FINISHED, ExtractionOrderState.FAILED]:
             self.state = get_order_status_from_conversion_progress(job_overall_progress)
--- a/osmaxx-py/osmaxx/excerptexport/models/extraction_order.py
+++ b/osmaxx-py/osmaxx/excerptexport/models/extraction_order.py
@@ -33,6 +33,16 @@ def get_order_status_from_conversion_progress(progress):
 
 
 class ExtractionOrder(models.Model):
+    DOWNLOAD_STATUS_UNKNOWN = 0
+    DOWNLOAD_STATUS_DOWNLOADING = 1
+    DOWNLOAD_STATUS_AVAILABLE = 2
+
+    DOWNLOAD_STATUSES = (
+        (DOWNLOAD_STATUS_UNKNOWN, 'unknown'),
+        (DOWNLOAD_STATUS_DOWNLOADING, 'downloading'),
+        (DOWNLOAD_STATUS_AVAILABLE, 'received'),
+    )
+
     state = enum.EnumField(ExtractionOrderState, default=ExtractionOrderState.INITIALIZED, verbose_name=_('state'))
     _extraction_configuration = models.TextField(
         blank=True, null=True, default='', verbose_name=_('extraction options')
@@ -41,6 +51,7 @@ class ExtractionOrder(models.Model):
     orderer = models.ForeignKey(User, related_name='extraction_orders', verbose_name=_('orderer'))
     excerpt = models.ForeignKey(Excerpt, related_name='extraction_orders', verbose_name=_('excerpt'))
     progress_url = models.URLField(verbose_name=_('progress URL'), null=True, blank=True)
+    download_status = models.IntegerField(_('file status'), choices=DOWNLOAD_STATUSES, default=DOWNLOAD_STATUS_UNKNOWN)
 
     def __str__(self):
         return '[' + str(self.id) + '] orderer: ' + self.orderer.get_username() + ', excerpt: ' + self.excerpt.name +\
--- a/osmaxx-py/osmaxx/excerptexport/models/extraction_order.py
+++ b/osmaxx-py/osmaxx/excerptexport/models/extraction_order.py
@@ -33,12 +33,12 @@ def get_order_status_from_conversion_progress(progress):
 
 
 class ExtractionOrder(models.Model):
-    DOWNLOAD_STATUS_UNKNOWN = 0
+    DOWNLOAD_STATUS_NOT_DOWNLOADED = 0
     DOWNLOAD_STATUS_DOWNLOADING = 1
     DOWNLOAD_STATUS_AVAILABLE = 2
 
     DOWNLOAD_STATUSES = (
-        (DOWNLOAD_STATUS_UNKNOWN, 'unknown'),
+        (DOWNLOAD_STATUS_NOT_DOWNLOADED, 'unknown'),
         (DOWNLOAD_STATUS_DOWNLOADING, 'downloading'),
         (DOWNLOAD_STATUS_AVAILABLE, 'received'),
     )
@@ -51,7 +51,11 @@ class ExtractionOrder(models.Model):
     orderer = models.ForeignKey(User, related_name='extraction_orders', verbose_name=_('orderer'))
     excerpt = models.ForeignKey(Excerpt, related_name='extraction_orders', verbose_name=_('excerpt'))
     progress_url = models.URLField(verbose_name=_('progress URL'), null=True, blank=True)
-    download_status = models.IntegerField(_('file status'), choices=DOWNLOAD_STATUSES, default=DOWNLOAD_STATUS_UNKNOWN)
+    download_status = models.IntegerField(
+        _('file status'),
+        choices=DOWNLOAD_STATUSES,
+        default=DOWNLOAD_STATUS_NOT_DOWNLOADED
+    )
 
     def __str__(self):
         return '[' + str(self.id) + '] orderer: ' + self.orderer.get_username() + ', excerpt: ' + self.excerpt.name +\
--- a/osmaxx-py/osmaxx/excerptexport/models/extraction_order.py
+++ b/osmaxx-py/osmaxx/excerptexport/models/extraction_order.py
@@ -82,4 +82,3 @@ class ExtractionOrder(models.Model):
         # don't update if we already we're done or errored out
         if self.state not in [ExtractionOrderState.FINISHED, ExtractionOrderState.FAILED]:
             self.state = get_order_status_from_conversion_progress(job_overall_progress)
-            self.save()
--- a/osmaxx-py/osmaxx/excerptexport/services/conversion_api_client.py
+++ b/osmaxx-py/osmaxx/excerptexport/services/conversion_api_client.py
@@ -99,14 +99,9 @@ class ConversionApiClient(RESTApiJWTClient):
             if extraction_order.download_status == extraction_order.DOWNLOAD_STATUS_UNKNOWN:
                 extraction_order.download_status = extraction_order.DOWNLOAD_STATUS_DOWNLOADING
                 extraction_order.save()
-                try:
-                    for download_file in job_status['gis_formats']:
-                        assert download_file['progress'] == 'successful'
-                        self._download_file(download_file, extraction_order)
-                finally:
-                    extraction_order.download_status = extraction_order.DOWNLOAD_STATUS_UNKNOWN
-                    extraction_order.save()
-
+                for download_file in job_status['gis_formats']:
+                    assert download_file['progress'] == 'successful'
+                    self._download_file(download_file, extraction_order)
                 extraction_order.download_status = extraction_order.DOWNLOAD_STATUS_AVAILABLE
                 extraction_order.save()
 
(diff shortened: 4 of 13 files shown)


docs/project-development-environment.md
```diff
@@ -55,7 +55,9 @@ To run the application tests only, see [Commonly used commands while developing
 
 ## Access the application
 
-[http://localhost:8000](http://localhost:8000)
+`http://<your_ip>:8000`
+
+where `<your_ip>` is your (public) IP.
```
wasabideveloper: Am I right: We are not longer able to run the application in a local network?

hixi: No, local network is not an issue, just localhost (loopback) isn't working anymore.

osmaxx-py/osmaxx/excerptexport/models/extraction_order.py
```diff
@@ -28,6 +51,7 @@ class ExtractionOrder(models.Model):
     orderer = models.ForeignKey(User, related_name='extraction_orders', verbose_name=_('orderer'))
     excerpt = models.ForeignKey(Excerpt, related_name='extraction_orders', verbose_name=_('excerpt'))
     progress_url = models.URLField(verbose_name=_('progress URL'), null=True, blank=True)
+    download_status = models.IntegerField(_('file status'), choices=DOWNLOAD_STATUSES, default=DOWNLOAD_STATUS_UNKNOWN)
```
wasabideveloper: Why do we not use an enum field like `state`?

hixi: Because it is common to do it like this in [Django](https://docs.djangoproject.com/en/1.9/ref/models/fields/#choices).

osmaxx-py/osmaxx/excerptexport/services/conversion_api_client.py
```diff
@@ -83,37 +85,44 @@ def create_job(self, extraction_order, callback_host):
...
-            return True
-        else:
-            return False
+        with transaction.atomic():
+            extraction_order.refresh_from_db()
+            if extraction_order.download_status == extraction_order.DOWNLOAD_STATUS_UNKNOWN:
+                extraction_order.download_status = extraction_order.DOWNLOAD_STATUS_DOWNLOADING
+                extraction_order.save()
+                success = True
+                for download_file in job_status['gis_formats']:
```
wasabideveloper: Do we download only the gis files?

hixi: What other files are there? We download every formats file, no matter if `gis` or not.

osmaxx-py/osmaxx/excerptexport/services/conversion_api_client.py
```diff
@@ -83,37 +85,44 @@ def create_job(self, extraction_order, callback_host):
...
+                    if download_file['progress'] == 'successful':
+                        self._download_file(download_file, extraction_order)
+                    else:
+                        success = False
+                if not success:
```
wasabideveloper: How are failed downloads handled? Try again later?

osmaxx-py/osmaxx/excerptexport/services/conversion_api_client.py
wasabideveloper: Garmin will be a GIS file?

osmaxx-py/osmaxx/excerptexport/models/extraction_order.py
wasabideveloper: I'd like to do it always the same. And if there is a enum libray, why should we simulate this by tupples by our own?

osmaxx-py/osmaxx/excerptexport/services/conversion_api_client.py
hixi: Then they are downloaded in the moment that the user accesses the result page.

osmaxx-py/osmaxx/excerptexport/services/conversion_api_client.py
hixi: Garmin isn't available currently. If it should be available, we will come up with a download solution for it as well. Will be part of #154

osmaxx-py/osmaxx/excerptexport/services/conversion_api_client.py
wasabideveloper: If you think other code readers need this information, please add a comment.
:heavy_check_mark:

osmaxx-py/osmaxx/excerptexport/services/conversion_api_client.py

osmaxx-py/osmaxx/excerptexport/models/extraction_order.py
hixi: Why use a special solution instead of best practices? I do like to do it the same everywhere, which means I'd rather change the `enum` we're using and switch to a more compliant way.

wasabideveloper: I see the problem we have using this special implementation. So I agree to remove the enum_field sometime.
:heavy_check_mark:

docs/project-development-environment.md
das-g: On Linux,
```bash
ip route get 1 | awk '{print $NF;exit}'
```
seems to be a pretty reliable way to determine the IP to use. (Found [on StackOverflow](http://stackoverflow.com/questions/13322485/how-to-i-get-the-primary-ip-address-of-the-local-machine-on-linux-and-os-x/25851186#25851186).)

If it answers `RTNETLINK answers: Network is unreachable` one will have to manually add an entry in `/etc/hosts` and use that.

Should any of this be mentioned here in the docu?

osmaxx-py/osmaxx/excerptexport/migrations/0012_outputfile_file_status.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ('excerptexport', '0011_remove_extractionorder_process_start_date'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='outputfile',
+            name='file_status',
+            field=models.IntegerField(verbose_name='file status', choices=[(1, 'downloading'), (2, 'received')], default=0),
```
das-g: Is it allowed to specify a `default` that isn't amongst the `choices`?

osmaxx-py/osmaxx/excerptexport/models/extraction_order.py
```diff
@@ -64,3 +88,8 @@ def extraction_configuration(self, value):
     @property
     def extraction_formats(self):
         return json.loads(self.extraction_configuration).get('gis_formats', None)
+
+    def set_status_from_conversion_progress(self, job_overall_progress):
+        # Don't update if we're already done or have errored out:
```
das-g: Is this comment needed? To me, what it says is mostly evident from the code itself.

osmaxx-py/osmaxx/excerptexport/services/conversion_api_client.py
```diff
@@ -83,37 +85,44 @@ def create_job(self, extraction_order, callback_host):
...
-            return False
+        with transaction.atomic():
+            extraction_order.refresh_from_db()
+            if extraction_order.download_status == extraction_order.DOWNLOAD_STATUS_UNKNOWN:
+                extraction_order.download_status = extraction_order.DOWNLOAD_STATUS_DOWNLOADING
+                extraction_order.save()
+                success = True
+                for download_file in job_status['gis_formats']:
+                    if download_file['progress'] == 'successful':
+                        self._download_file(download_file, extraction_order)
```
das-g: Our current naming makes this condition confusing. It downloads successfully generated (and thus available on the conversion service) files. However, it reads somewhat as if already successfully downloaded files are being downloaded.

osmaxx-py/osmaxx/excerptexport/services/conversion_api_client.py
das-g: Consider handling the positive case in the `if` and the negative one in the `else`.

osmaxx-py/osmaxx/excerptexport/services/conversion_api_client.py
```diff
@@ -83,37 +85,44 @@ def create_job(self, extraction_order, callback_host):
...
+                    else:
+                        success = False
+                if not success:
+                    extraction_order.download_status = extraction_order.DOWNLOAD_STATUS_UNKNOWN
+                    extraction_order.save()
+                else:
+                    extraction_order.download_status = extraction_order.DOWNLOAD_STATUS_AVAILABLE
+                    extraction_order.save()
```
das-g: Should the common `extraction_order.save()` be pulled out of the `if`-`else`?

osmaxx-py/osmaxx/excerptexport/services/conversion_api_client.py
```diff
@@ -132,12 +141,12 @@ def job_status(self, extraction_order):
                         {
                             "format": "fgdb",
                             "progress": "successful",
-                            "result_url": "http://localhost:8000/api/gis_format/11/download_result/"
+                            "result_url": "http://<your_ip>:8000/api/gis_format/11/download_result/"
```
das-g: It's just an example, but as we're changing it here, anyway, shouldn't we change it to something like `http://<conversion service host>:8901/api/gis_format/11/download_result/`?

osmaxx-py/osmaxx/excerptexport/services/conversion_api_client.py
```diff
@@ -168,18 +176,11 @@ def update_order_status(self, extraction_order):
...
-                extraction_order.state = ExtractionOrderState.PROCESSING
-                extraction_order.save()
-            return True
-        else:
-            return False
+            progress = job_status['progress']
+            extraction_order.set_status_from_conversion_progress(progress)
+            extraction_order.save()
+            if progress == 'successful':
+                self._download_result_files(extraction_order, job_status)
```
das-g: This change turns the docstring wrong. (Function doesn't return anything anymore.)

docs/project-development-environment.md
hixi: If you think this is something worth adding, add it :-)

osmaxx-py/osmaxx/excerptexport/tests/test_conversion_api_client.py
```diff
@@ -120,15 +121,17 @@ def test_download_files(self):
...
 
             if cassette_empty:
                 # wait for external service to complete request
                 time.sleep(120)
 
-            success = self.api_client.download_result_files(self.extraction_order)
+            self.api_client._download_result_files(
+                self.extraction_order,
+                job_status=self.api_client.job_status(self.extraction_order)
+            )
```
das-g: As `_download_result_files` is now private, should it still be tested directly or only through `update_order_status`?

osmaxx-py/osmaxx/excerptexport/services/conversion_api_client.py
hixi: good catch

osmaxx-py/osmaxx/utilities/shortcuts.py
```diff
@@ -32,3 +38,44 @@ def get_list_or_none(klass, *args, **kwargs):
     if not object_list:
         return None
     return object_list
+
+
+class Emissary:
+    def __init__(self, user):
```
das-g: Name the constructor arguments (and the corresponding fields) according to the roles they have w.r.t. the created emissary object rather than their type. (E.g. `recepient` instead of `user`.)

Otherwise construction looks like an Emissary is a special user that wraps a normal user.

osmaxx-py/osmaxx/excerptexport/tests/test_conversion_api_client.py
hixi: I think we can leave it as is, if it stands in our way anytime later, we may change it then.

osmaxx-py/osmaxx/job_progress/views.py
```diff
@@ -1,11 +1,48 @@
...
+        finished_email_body = _(
+            'The extraction order "{order_id}" could not be completed, please try again later.'
+        ).format(order_id=order.id)
+
+        emissary.error(message)
+        emissary.inform_mail(subject=finished_email_subject, mail_body=finished_email_body)
+    else:
+        message = _('The progress of the extraction order "{order_id}" has been updated.').format(
+            order_id=order.id,
+        )
```
das-g: The user doesn't care about the progress being information distributed between our components (webapp and conversion service). Thus 'updated' might be confusing.

How about something like `"Extraction order {order} is now {order_state}.".format(order=order.id, order_state=str(order.state))`?

osmaxx-py/osmaxx/utilities/shortcuts.py
```diff
@@ -32,3 +38,44 @@ def get_list_or_none(klass, *args, **kwargs):
...
+        self.inform(messages.WARNING, message)
+
+    def error(self, message):
+        self.inform(messages.ERROR, message)
+
+    def debug(self, message):
+        self.inform(messages.DEBUG, message)
+
+    def inform_mail(self, subject, mail_body, warn_if_no_email=True):
+        email = getattr(self.user, 'email', None)
```
das-g: Consider calling the variable `email_address`.

osmaxx-py/osmaxx/utilities/shortcuts.py
```diff
@@ -32,3 +38,44 @@ def get_list_or_none(klass, *args, **kwargs):
...
+        if email:
```
das-g: Candidate for [EAPF](https://docs.python.org/3/glossary.html#term-eafp).

osmaxx-py/osmaxx/utilities/shortcuts.py
```diff
@@ -32,3 +38,44 @@ def get_list_or_none(klass, *args, **kwargs):
...
+            mail.send_mail(
+                '[OSMAXX] ' + subject,
+                mail_body,
+                settings.DEFAULT_FROM_EMAIL,
+                [self.user.email]
+            )
+        if warn_if_no_email and not email:
```
das-g: If not switching to EAPF, consider using `elif warn_if_no_email:`

osmaxx-py/osmaxx/utilities/shortcuts.py
```diff
@@ -32,3 +38,44 @@ def get_list_or_none(klass, *args, **kwargs):
...
+        self.inform(messages.DEBUG, message)
+
+    def inform_mail(self, subject, mail_body, warn_if_no_email=True):
+        email = getattr(self.user, 'email', None)
+        if email:
+            mail.send_mail(
+                '[OSMAXX] ' + subject,
+                mail_body,
+                settings.DEFAULT_FROM_EMAIL,
+                [self.user.email]
```
das-g: Why not use the local variable you already have?

osmaxx-py/osmaxx/utilities/shortcuts.py
das-g: :white_check_mark: fixed with 33ad8e5

osmaxx-py/osmaxx/job_progress/views.py
das-g: :white_check_mark: changed with 53ecc8f

osmaxx-py/osmaxx/excerptexport/models/extraction_order.py
das-g: :white_check_mark: removed with 6c6e900

osmaxx-py/osmaxx/excerptexport/services/conversion_api_client.py
das-g: :white_check_mark: corrected with 61a0db4

osmaxx-py/osmaxx/excerptexport/services/conversion_api_client.py
das-g: :white_check_mark: changed with abd8803

osmaxx-py/osmaxx/excerptexport/tests/test_conversion_api_client.py
das-g: :white_check_mark: Agreed, leave as-is.

docs/project-development-environment.md
das-g: :white_check_mark: done: 53a9ce6

osmaxx-py/osmaxx/excerptexport/services/conversion_api_client.py
```diff
@@ -83,37 +85,37 @@ def create_job(self, extraction_order, callback_host):
...
-        else:
-            return False
+        with transaction.atomic():
+            extraction_order.refresh_from_db()
+            if extraction_order.download_status == extraction_order.DOWNLOAD_STATUS_UNKNOWN:
+                extraction_order.download_status = extraction_order.DOWNLOAD_STATUS_DOWNLOADING
+                extraction_order.save()
+                for download_file in job_status['gis_formats']:
+                    assert download_file['progress'] == 'successful'
+                    self._download_file(download_file, extraction_order)
```
das-g: Our current naming makes this `if`-condition confusing. It downloads successfully generated (and thus available on the conversion service) files. However, it reads somewhat as if already successfully downloaded files are being downloaded (again).

osmaxx-py/osmaxx/excerptexport/services/conversion_api_client.py
das-g: :white_check_mark: obsoleted by c21e5cb & 9f41bbd

osmaxx-py/osmaxx/utilities/shortcuts.py
```diff
@@ -32,3 +38,44 @@ def get_list_or_none(klass, *args, **kwargs):
...
+        self.inform(messages.WARNING, message)
+
+    def error(self, message):
+        self.inform(messages.ERROR, message)
+
+    def debug(self, message):
+        self.inform(messages.DEBUG, message)
+
+    def inform_mail(self, subject, mail_body, warn_if_no_email=True):
+        email = getattr(self.recipient, 'email', None)
```
das-g: Consider calling the variable `email_address`.

osmaxx-py/osmaxx/utilities/shortcuts.py
```diff
@@ -32,3 +38,44 @@ def get_list_or_none(klass, *args, **kwargs):
...
+        if email:
```
das-g: Candidate for [EAPF](https://docs.python.org/3/glossary.html#term-eafp).

osmaxx-py/osmaxx/utilities/shortcuts.py
```diff
@@ -32,3 +38,44 @@ def get_list_or_none(klass, *args, **kwargs):
...
+            mail.send_mail(
+                '[OSMAXX] ' + subject,
+                mail_body,
+                settings.DEFAULT_FROM_EMAIL,
+                [self.recipient.email]
+            )
+        if warn_if_no_email and not email:
```
das-g: If not switching to EAPF, consider using `elif warn_if_no_email:`

osmaxx-py/osmaxx/utilities/shortcuts.py
```diff
@@ -32,3 +38,44 @@ def get_list_or_none(klass, *args, **kwargs):
...
+        self.inform(messages.DEBUG, message)
+
+    def inform_mail(self, subject, mail_body, warn_if_no_email=True):
+        email = getattr(self.recipient, 'email', None)
+        if email:
+            mail.send_mail(
+                '[OSMAXX] ' + subject,
+                mail_body,
+                settings.DEFAULT_FROM_EMAIL,
+                [self.recipient.email]
```
das-g: Why not use the local variable you already have?

Issue #293, opened 2015-12-08
Translate JS strings in Frontend


wasabideveloper:

hixi:
Resolution possibility: Remove all hardcoded labels from js and use templetized divs or spans.

wasabideveloper:
Maybe it's a good idea to use data-* attributes and read the labels in JS from there

das-g:
Django seems to have a built-in solution for exposing translations to JavaScript: https://docs.djangoproject.com/en/1.10/topics/i18n/translation/#internationalization-in-javascript-code

Issue #254, opened 2015-11-09
Fix usability issues part2


wasabideveloper:
Estimated duration: ~0.5d

Part 1: #199 
Usability test result: #142

- [x] Add count to select category
- [ ] Add status preparing

wasabideveloper:
"Add status preparing" will be fixed by an other implementation.

PR #294, opened 2015-12-08
Feature/client validation#19

### LINKED ISSUE #19: Implement frontend (client) validation
* check for same as on server: #20 
* in case of constraint violation:
  * indicate cause (highlighting of widgets with invalid stuff / display hints / ...)
  * disable request creation button until all validations pass again

wasabideveloper:
Implementation of feature #19 (Merge will close #19 ).

* JS string **NOT** localized. Todo later: #293 .
* ran tests
* tested views by hand:
  * /orders/new [get/post]

To be reviewed by:
- [ ] @das-g 
- [x] @hixi 

*Contains changes of #254 *

<hr />
screen shot 2015-12-08 at 15 28 39
</hr>


osmaxx-py/osmaxx/excerptexport/services/conversion_api_client.py
```diff
@@ -181,6 +182,21 @@ def update_order_status(self, extraction_order):
         else:
             return False
 
+    def estimated_file_size(self, north=0, west=0, south=0, east=0):
```
das-g: Is it a good idea to have default argument values for this method?

osmaxx-py/osmaxx/excerptexport/static/excerptexport/scripts/formValidation.js
```diff
@@ -32,15 +66,60 @@
...
+                    var estimatedFileSize = Number(data['estimated_file_size_in_bytes']);
+                    this.validity['excerptBounds'] = estimatedFileSize < allowedMaxSize;
+                    this.setSubmitButtonState();
+                    
+                    this.excerptBoundInputFields.forEach(function(excerptBoundInputField) {
+                        if(this.validity['excerptBounds']) {
+                            excerptBoundInputField.setCustomValidity('');
+                            document.getElementById('excerpt-validation').textContent = '';
+                        } else {
+                            var howMuchTooLarge = estimatedFileSize ? Math.round(100 / allowedMaxSize * estimatedFileSize - 100) + '% ': '';
```
das-g: Would `Math.ceil(...)` be better suited here? I don't think we want confusing `Excerpt 0% too large!` messages.

das-g: The computation here would be more readable with the multiplication done before the division, I think.

osmaxx-py/osmaxx/excerptexport/services/conversion_api_client.py
wasabideveloper: Is it possible to get named arguments without default arguments? I don't like positional arguments for coordinates because every one uses an other ordering, so it would be much error-prone.

osmaxx-py/osmaxx/excerptexport/static/excerptexport/scripts/formValidation.js
wasabideveloper: :heavy_check_mark:

osmaxx-py/osmaxx/excerptexport/services/conversion_api_client.py
das-g: Yes, keyword arguments and defaults can be used separately. In fact, every argument can be used as a keyword argument at the caller's discretion, whether it has a default or not.

PR #304, opened 2015-12-11
Feature/update status on failed conversion#288

### LINKED ISSUE #288: webapp: set status on failed if process never finishes/returns
If a conversion jub doesn't finish (eg. stays on `processing`), set the job to error on the osmaxx side.

`x` stands for a day or something similar:
- [ ] after `x`, set status to error, if not finished
- [ ] log as an error so we can see it in sentry

wasabideveloper:
Implementation of feature #288 (Merge will close #288 ).

* ran tests
* tested views by hand:
  * /orders/new [get/post]
  * /orders/{id}
  * /orders

To be reviewed by:
- [x] @das-g 
- [x] @hixi

### DIFF
--- a/osmaxx-py/config/settings/common.py
+++ b/osmaxx-py/config/settings/common.py
@@ -11,8 +11,9 @@ https://docs.djangoproject.com/en/1.8/ref/settings/
 import os
 from datetime import timedelta
 import environ
-import sys
+
 from django.contrib.messages import constants as message_constants
+from django.utils import timezone
 
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
 BASE_DIR = os.path.dirname(os.path.dirname(__file__))
@@ -375,6 +376,9 @@ OSMAXX = {
     'download_file_name': '%(date)s-%(excerpt_name)s-%(id)s.%(content_type)s.%(file_extension)s',
     'download_chunk_size': 8192,
     'orders_history_number_of_items': 100,
+    'EXTRACTION_PROCESSING_TIMEOUT_TIMEDELTA': timezone.timedelta(
+        hours=env.str('DJANGO_OSMAXX_EXTRACTION_PROCESSING_TIMEOUT_HOURS', default=12)
+    ),
     # The email adress of this user will be used to generate the mailto link for users
     # to request access to osmaxx (access_denied page)
     'ACCOUNT_MANAGER_EMAIL': env.str('OSMAXX_ACCOUNT_MANAGER_EMAIL', default=DEFAULT_FROM_EMAIL),
--- a/osmaxx-py/osmaxx/excerptexport/models/extraction_order.py
+++ b/osmaxx-py/osmaxx/excerptexport/models/extraction_order.py
@@ -103,6 +103,11 @@ class ExtractionOrder(models.Model):
     def extraction_formats(self):
         return json.loads(self.extraction_configuration).get('gis_formats', None)
 
+    @property
+    def process_due_time(self):
+        from django.conf import settings  # import locally, so migrations can't depend on settings
+        return self.process_start_time + settings.OSMAXX.get('EXTRACTION_PROCESSING_TIMEOUT_TIMEDELTA')
+
     def set_status_from_conversion_progress(self, job_overall_progress):
         if self.state not in [ExtractionOrderState.FINISHED, ExtractionOrderState.FAILED]:
             self.state = get_order_status_from_conversion_progress(job_overall_progress)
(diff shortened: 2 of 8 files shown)


osmaxx-py/osmaxx/excerptexport/services/conversion_api_client.py
```diff
@@ -22,6 +23,12 @@ class ConversionApiClient(RESTApiJWTClient):
     conversion_job_url = '/jobs/'
     conversion_job_status_url = '/conversion_result/{job_uuid}/'
 
+    @staticmethod
+    def _extraction_processing_overdated(progress, extraction_order):
+        return (progress in ['new', 'received', 'started']) and timezone.now() > (
+            extraction_order.process_start_time + settings.OSMAXX.get('EXTRACTION_PROCESSING_TIMEOUT_TIMEDELTA')
```
hixi: `process_start_time` can be explicitly null (None), which gives an error whenever adding with `timedelta`. Maybe checking if `extraction_order.process_start_time` is not None?

osmaxx-py/osmaxx/excerptexport/models/extraction_order.py
```diff
@@ -51,6 +51,7 @@ class ExtractionOrder(models.Model):
     orderer = models.ForeignKey(User, related_name='extraction_orders', verbose_name=_('orderer'))
     excerpt = models.ForeignKey(Excerpt, related_name='extraction_orders', verbose_name=_('excerpt'))
     progress_url = models.URLField(verbose_name=_('progress URL'), null=True, blank=True)
+    process_start_time = models.DateTimeField(verbose_name=_('process start time'), null=True, blank=True)
```
hixi: instead of blank, we could also use [auto_now_add](https://docs.djangoproject.com/en/1.9/ref/models/fields/#django.db.models.DateField.auto_now_add) - then it wont be set from inside a view, but automatically, which might or might not be the desired behavior.

osmaxx-py/osmaxx/excerptexport/services/conversion_api_client.py

osmaxx-py/config/settings/common.py
```diff
@@ -375,6 +376,9 @@
     'download_file_name': '%(date)s-%(excerpt_name)s-%(id)s.%(content_type)s.%(file_extension)s',
     'download_chunk_size': 8192,
     'orders_history_number_of_items': 100,
+    'EXTRACTION_PROCESSING_TIMEOUT_TIMEDELTA': timezone.timedelta(
+        hours=env.str('DJANGO_OSMAXX_EXTRACTION_PROCESSING_TIMEOUT_HOURS', default=12)
```
hixi: I think `env.int` (instead of `env.str`) is what is intended here, because otherwise the conversion from a environment variable isn't handled correctly.

osmaxx-py/osmaxx/excerptexport/services/conversion_api_client.py
```diff
@@ -29,6 +30,12 @@ class ConversionApiClient(RESTApiJWTClient):
     estimated_file_size_url = '/estimate_size_in_bytes/'
     country_base_url = '/country/'
 
+    @staticmethod
+    def _extraction_processing_overdated(progress, extraction_order):
```
das-g: Did you mean 'overdue'?

osmaxx-py/osmaxx/excerptexport/services/conversion_api_client.py
```diff
@@ -29,6 +30,12 @@ class ConversionApiClient(RESTApiJWTClient):
...
+        return extraction_order.process_start_time and (progress in ['new', 'received', 'started']) and timezone.now() > (
+            extraction_order.process_start_time + settings.OSMAXX.get('EXTRACTION_PROCESSING_TIMEOUT_TIMEDELTA')
+        )
```
das-g: Giving intermediate results a name (by extracting them to local variables or to functions) might make this logic easier to understand. Candidates for extracting:
* `progress in ['new', 'received', 'started']` (process unfinished)
* `timezone.now() > (extraction_order.process_start_time + settings.OSMAXX.get('EXTRACTION_PROCESSING_TIMEOUT_TIMEDELTA'))` (timeout reached)
* `extraction_order.process_start_time + settings.OSMAXX.get('EXTRACTION_PROCESSING_TIMEOUT_TIMEDELTA')` (process due time)

osmaxx-py/osmaxx/excerptexport/services/conversion_api_client.py

osmaxx-py/osmaxx/excerptexport/services/conversion_api_client.py

PR #307, opened 2015-12-11
quick fix for sending with user email as sender

### LINKED ISSUE #298: Don't send emails on behalf of FROM addresses we don't own
We [are sending emails on behalf of the user](https://github.com/geometalab/osmaxx/blob/49ce929de03b6a696fb8e4ea2f4c0b30d219b911/osmaxx-py/osmaxx/excerptexport/views.py#L175) (i.e., with their email address as `FROM` header). We send these emails to ourselves (thus urgency low), but still:
~~If we want to continue to send email on behalf of addresses we don't own, [we should at least also set the `SENDER` header](http://stackoverflow.com/a/18906678/674064) to an address we _do_ own. (And for which our server is seen as a legitimate email message source (tracked in #300).)~~
We should use a `FROM` address we own for sending email.

hixi:
closes #298 

Reviewed by:
- [ ] @das-g

### DIFF
--- a/osmaxx-py/osmaxx/excerptexport/views.py
+++ b/osmaxx-py/osmaxx/excerptexport/views.py
@@ -158,15 +158,18 @@ def request_access(request):
         )
     else:
         email_message = _(
-            '{first_name} {last_name} requests access for Osmaxx. Please activate the user {username}.'
+            '''Hi Admin!
+            {first_name} {last_name} ({email}) requests access for Osmaxx. Please activate the user {username}.
+            '''
         ).format(
             username=request.user.username,
             first_name=request.user.first_name,
-            last_name=request.user.last_name
+            last_name=request.user.last_name,
+            email=request.user.email,
         )
 
         try:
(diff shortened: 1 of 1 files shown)


osmaxx-py/osmaxx/excerptexport/views.py
```diff
@@ -158,15 +158,18 @@ def request_access(request):
...
             username=request.user.username,
             first_name=request.user.first_name,
-            last_name=request.user.last_name
+            last_name=request.user.last_name,
+            email=request.user.email,
         )
 
         try:
-            send_mail('Request access for Osmaxx', email_message, request.user.email,
+            send_mail('Request access for Osmaxx', email_message, user_administrator_email,
```
das-g: Why not use Django's DEFAULT_FROM?

hixi: Good idea, was a bit too quick with the fix...

PR #317, opened 2015-12-14
Feature/return geo json feature wrapper

hixi:
Adds possibility to distinguish between countries and boundingboxexcerpts on the JavaScript side.

Reviewed by:

- [x] @wasabideveloper 
- [x] @das-g 

This should also be a basis, so that validation can be turned on again.

### DIFF
--- a/osmaxx-py/osmaxx/excerptexport/static/excerptexport/scripts/map.js
+++ b/osmaxx-py/osmaxx/excerptexport/static/excerptexport/scripts/map.js
@@ -6,6 +6,7 @@
         this.selectElementExistingExcerpts = selectElementExistingExcerpts;
         this.formElementPartsSwitcher = formElementPartsSwitcher;
         this.selectedExcerptGeoJson = null;
+        this.country = null;
 
         /**
          * Synchronize coordinates in input fields to excerpt on map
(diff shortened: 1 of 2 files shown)


osmaxx-py/osmaxx/excerptexport/static/excerptexport/scripts/map.js
```diff
@@ -88,12 +89,39 @@
...
-                map.fitBounds(that.locationFilter.getBounds());
+                switch(feature_type) {
+                    case 'BBoxBoundingGeometry':
+                        that._handleBBoxBoundingGeometry(this);
+                        break;
+                    case 'Country':
+                        that._handleCountry(this);
+                        break;
+                    default:
+                        break;
```
das-g: Should we
```js
console.warn("Can't visualize excerpt with unexpected geometry type '" + feature_type + "': " + this.getLayers()[0])
```
(or similar) or even `console.error(...)` in the `default` case?

hixi: Since we don't get the error to an error logger like sentry in production[1], I think this doesn't make much sense, we already see (as developers), when `nothing` happens...

[1]: though we _[could](https://docs.getsentry.com/on-premise/clients/javascript/)_ if we really wanted to ;-)

das-g: > [1]: though we _[could](https://docs.getsentry.com/on-premise/clients/javascript/)_ if we really wanted to ;-)

Heh, nice. Good to keep in mind for cases where it's more necessary than here.

> Since we don't get the error to an error logger like sentry in production[1], I think this doesn't make much sense, we already see (as developers), when `nothing` happens...

Fair enough. :white_check_mark: Leave as-is.

PR #356, opened 2015-12-23
Feature/log timeout failed#348

### LINKED ISSUE #348: Log order status set to failed cause of timeout
If the order takes to long, the webapp will set the status to failed. This need to be logged.

use level warning and log level warning.

wasabideveloper:
Fix of bug #348 (Merge will close #348 ).

* ran tests

To be reviewed by:
- [x] @das-g 
- [x] @hixi

### DIFF
--- a/web_frontend/osmaxx/excerptexport/services/conversion_api_client.py
+++ b/web_frontend/osmaxx/excerptexport/services/conversion_api_client.py
@@ -191,6 +191,12 @@ class ConversionApiClient(RESTApiJWTClient):
             if progress == 'successful':
                 self._download_result_files(extraction_order, job_status)
             elif self._extraction_processing_overdue(progress, extraction_order):
+                logger.warning(
+                    'Extraction order %s processing timeout overdue. Set status to FAILED. %s, %s',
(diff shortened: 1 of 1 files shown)


web_frontend/osmaxx/excerptexport/services/conversion_api_client.py
```diff
@@ -191,6 +192,12 @@ def update_order_status(self, extraction_order):
             if progress == 'successful':
                 self._download_result_files(extraction_order, job_status)
             elif self._extraction_processing_overdue(progress, extraction_order):
+                logger.warning(
+                    'Extraction order %s processing timeout overdue. Set status to FAILED. %s, %s',
+                    extraction_order.id,
+                    serializers.serialize('json', extraction_order),
```
hixi: why would you want to log the entire extraction order in json? An ID and name does suffice, as we can always track this in the admin to see the details (using that ID).

wasabideveloper: :heavy_check_mark:

PR #361, opened 2016-01-07
refactoring: extract common form_valid() method of OrderNewExcerptView and OrderExistingExcerptView

das-g:
I've decided to let the new mixin inherit from `FormMixin`, which provides the overridden method. MRO looks OK, and I don't think the resolution order is particularly fragile for this case.

reviewed by:
- [x] @hixi

### DIFF
--- a/web_frontend/osmaxx/excerptexport/views.py
+++ b/web_frontend/osmaxx/excerptexport/views.py
@@ -11,6 +11,7 @@ from django.contrib import messages
 from django.utils.translation import ugettext_lazy as _
 from django.views.generic import FormView
 from django.conf import settings
+from django.views.generic.edit import FormMixin
 
 from osmaxx.excerptexport.services.shortcuts import get_authenticated_api_client
 from .models import ExtractionOrder, OutputFile
@@ -31,10 +32,7 @@ def execute_converters(extraction_order, callback_host, protocol):
     get_authenticated_api_client().create_job(extraction_order, callback_host=callback_host, protocol=protocol)
 
 
-class OrderNewExcerptView(LoginRequiredMixin, FrontendAccessRequiredMixin, FormView):
-    template_name = 'excerptexport/templates/order_new_excerpt.html'
-    form_class = ExcerptForm
-
+class OrderFormViewMixin(FormMixin):
     def form_valid(self, form):
         extraction_order = form.save(self.request.user)
         request_host = self.request.get_host()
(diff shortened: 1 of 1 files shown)


web_frontend/osmaxx/excerptexport/views.py
```diff
@@ -31,10 +32,7 @@ def execute_converters(extraction_order, callback_host, protocol):
     get_authenticated_api_client().create_job(extraction_order, callback_host=callback_host, protocol=protocol)
 
 
-class OrderNewExcerptView(LoginRequiredMixin, FrontendAccessRequiredMixin, FormView):
-    template_name = 'excerptexport/templates/order_new_excerpt.html'
-    form_class = ExcerptForm
-
+class OrderFormViewMixin(FormMixin):
```
hixi: what do we gain by inheriting? `class OrderFormViewMixin:` would have the same effect, especially since there is no call to super.

das-g: It ensures that no unwanted MRO with `FormMixin` before `OrderFormViewMixin` (so that our override of `form_valid` would be without effect) occurs due to improper inheritance in a derived class.

Instead, one would get a
> TypeError: Cannot create a consistent method resolution
> order (MRO) for bases object, FormMixin, OrderFormViewMixin

which should be enough to hint at the problem. You can test that this happens with the (admittedly conceived) change
```diff
diff --git a/web_frontend/osmaxx/excerptexport/views.py b/web_frontend/osmaxx/excerptexport/views.py
index fc6a0c0..11dd9fe 100644
--- a/web_frontend/osmaxx/excerptexport/views.py
+++ b/web_frontend/osmaxx/excerptexport/views.py
@@ -49,7 +49,7 @@ class OrderFormViewMixin(FormMixin):
         )
 
 
-class OrderNewExcerptView(LoginRequiredMixin, FrontendAccessRequiredMixin, OrderFormViewMixin, FormView):
+class OrderNewExcerptView(LoginRequiredMixin, FrontendAccessRequiredMixin, FormView, FormMixin, OrderFormViewMixin):
     template_name = 'excerptexport/templates/order_new_excerpt.html'
     form_class = ExcerptForm
 
```
However with
```diff
diff --git a/web_frontend/osmaxx/excerptexport/views.py b/web_frontend/osmaxx/excerptexport/views.py
index fc6a0c0..d32231a 100644
--- a/web_frontend/osmaxx/excerptexport/views.py
+++ b/web_frontend/osmaxx/excerptexport/views.py
@@ -32,7 +32,7 @@ def execute_converters(extraction_order, callback_host, protocol):
     get_authenticated_api_client().create_job(extraction_order, callback_host=callback_host, protocol=protocol)
 
 
-class OrderFormViewMixin(FormMixin):
+class OrderFormViewMixin(object):
     def form_valid(self, form):
         extraction_order = form.save(self.request.user)
         request_host = self.request.get_host()
@@ -49,7 +49,7 @@ class OrderFormViewMixin(FormMixin):
         )
 
 
-class OrderNewExcerptView(LoginRequiredMixin, FrontendAccessRequiredMixin, OrderFormViewMixin, FormView):
+class OrderNewExcerptView(LoginRequiredMixin, FrontendAccessRequiredMixin, FormView, FormMixin, OrderFormViewMixin):
     template_name = 'excerptexport/templates/order_new_excerpt.html'
     form_class = ExcerptForm
 
```
you get failing tests, but no hints on why those are failing.

hixi: :white_check_mark:

PR #388, opened 2016-01-27
Feature/#384 order status update in middleware

### LINKED ISSUE #384: update order stati in middleware
instead of in views where it can be forgotten

das-g:
closes #384

#### reviewed by:
- [x] @hixi


web_frontend/osmaxx/job_progress/middleware.py
```diff
@@ -0,0 +1,39 @@
...
+    current_user = request.user
+    if current_user.is_anonymous():
+        return
+    for order in ExtractionOrder.objects.exclude(state__in=FINAL_STATES).filter(orderer=current_user):
+        update_order_if_stale(order)
+
+
+def update_order_if_stale(extraction_order):
+    get_cached_or_set(
+        'extraction_order_{}_progress'.format(extraction_order.id),
```
das-g: Alternatively, we could use `extraction_order.progress_url` as the cache key. What do you think @hixi?

hixi: I like the id better, progress_url can IIRC change when updated.

hixi:
Can we change the local settings to using the in-memory cache as well (opposed to using dummy cache)?

das-g:
> Can we change the local settings to using the in-memory cache as well (opposed to using dummy cache)?

"Can"? I say we _must_.

PR #390, opened 2016-01-28
Feature/eliminate are_all_keys_in_dict() helper function

das-g:
A little refactoring.

#### Reviewed by:
- [ ] @hixi


web_frontend/osmaxx/excerptexport/models/bounding_geometry.py
```diff
@@ -47,9 +46,9 @@ def __str__(self):
 
 class BBoxBoundingGeometry(BoundingGeometry):
     def __init__(self, *args, **kwargs):
-        attribute_names = ['north', 'east', 'south', 'west']
-        if are_all_keys_in(kwargs, attribute_names) \
-                and not are_all_keys_in(kwargs, ['south_west', 'north_east']):
+        attribute_names = {'north', 'east', 'south', 'west'}
+        if set(attribute_names).issubset(kwargs) \
+                and not {'south_west', 'north_east'}.issubset(kwargs):
```
das-g: The logic here is rather dubious, but I wanted to keep this PR a pure refactoring, so I didn't change it.

hixi: IMO, this make understanding it more difficult, can you elaborate on the reason for changing this?

das-g: This uses Python builtins whose semantics can be looked up, so we don't have to document them.

While the `are_all_key_in()` function has a very speaking name, the fact that it has to take two arguments makes it not-so-easy (also not incredibly hard, granted) to understand.

If we opt to keep the helper function around, I'd still advice to apply 712ca1ea7e321a8ae68ef7602635c180320dbf36 and d3597857133eb8d988879aab5bdcb1fa705354b3.

hixi: I like the tests, no argument from me there. For me, just reading through the code, `are_all_keys_in` with `arg1`=my dictionary and `arg2`=my list of keys to test against makes it faster to understand what goes on than when using the inline version. I'm also not arguing against the implementation itself, I like that much better, but I would keep the encapsulation in a separate function.

das-g:
decided to keep the are_all_keys_in_dict() helper function

closing without merge

PR #394, opened 2016-02-01
Feature/#393 log whether order status was re-fetched by custom middleware

### LINKED ISSUE #393: log whether order status was re-fetched by our custom middleware
loglevel info

das-g:
closes #393 

#### reviewd by:
- [x] @hixi

### DIFF
--- a/web_frontend/osmaxx/job_progress/middleware.py
+++ b/web_frontend/osmaxx/job_progress/middleware.py
@@ -41,15 +41,18 @@ def update_order(extraction_order):
     _log_cache_miss(extraction_order)
     conversion_client = get_authenticated_api_client()
     conversion_client.update_order_status(extraction_order)
-    message = "Fetched, updated and cached ExtractionOrder {extraction_order} progress: {progress}".format(
-        extraction_order=extraction_order.id,
-        progress=ExtractionOrderState.label(extraction_order.state),
-    )
-    logger.debug(message)
+    if logger.isEnabledFor(logging.DEBUG):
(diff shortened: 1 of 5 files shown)


web_frontend/osmaxx/job_progress/middleware.py
```diff
@@ -29,11 +32,41 @@ def update_order_if_stale(extraction_order):
...
     return extraction_order.state
+
+
+def _log_cache_miss(extraction_order):
+    if not logger.isEnabledFor(logging.INFO):
+        return
+    message = "Cache miss - Re-fetching progress of pending ExtractionOrder {extraction_order}.".format(
+        extraction_order=extraction_order.id,
+    )
+    logger.info(message)
```
hixi: Is `info` the right level? `debug` seems more appropriate for me here, because the only time I want these messages to _pollute_ my console is during bug-hunting. Is there a special reason for the info level?

das-g: I chose `info`, because this is conceptually a mid- to high-level message (something happened or something else) and because we're doing something that might be unusual in the middleware. Thus when we have a reason to view the logs, it's probably something we'd want to be aware of. Someone who only wants to get output for potential problems should set loglevel `warning` and won't see these.

hixi: I do have a different opinion on this, but I don't think this matters enough to have a lengthy discussion about it. I'm OK with leaving it as is.

Issue #397, opened 2016-02-09
datetime.timedelta.seconds is not what you think it is


das-g:
A `timedelta` object represents a duration as `days`, `seconds` and `microseconds`. To convert a `timedelta` object to a duration in seconds, use its `total_seconds()` method.

Not an issue with our current use, as our durations happen to have sub-day length and don't require sub-second resolution. Should be fixed anyway, though.

das-g:
Fixed with #398

PR #398, opened 2016-02-09
fix timedelta-to-seconds conversion #397

das-g:
#### Reviewed by:
- [x] @hixi

### DIFF
--- a/web_frontend/osmaxx/utilities/shortcuts.py
+++ b/web_frontend/osmaxx/utilities/shortcuts.py
@@ -86,7 +86,8 @@ class Emissary:
 
 
 def get_cached_or_set(
-        cache_string, func, *args, timeout=datetime.timedelta(minutes=15).seconds, on_cache_hit=None, **kwargs):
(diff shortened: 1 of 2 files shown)


web_frontend/osmaxx/utilities/shortcuts.py
```diff
@@ -86,7 +86,8 @@ def inform(self, message_type, message):
 
 
 def get_cached_or_set(
-        cache_string, func, *args, timeout=datetime.timedelta(minutes=15).seconds, on_cache_hit=None, **kwargs):
+        cache_string, func, *args, timeout=datetime.timedelta(minutes=15).total_seconds(), on_cache_hit=None, **kwargs
+):
```
das-g: I wonder: Should we (in a separate change) allow `get_cached_or_set` to accept a `timedelta` object as `timeout`?

hixi: I think that would be a good idea.

PR #403, opened 2016-02-17
#381 refactor mailto links

### LINKED ISSUE #381: refactor templates with mailto: links
In #373 I touched two partials that create mailto: links with subject and body. The creation of those links would benefit from being extracted to a separate, new partial:
- The templates currently creating those links semi-hardcoded would become cleaner
- DRY
- It would force us to move the generation of the link content (which will be the context of the new partial) to python, leading to better separation. (Python might in turn generate the content by rendering yet another template, if suitable. That might then be reused in #371)

das-g:
closes #381

#### reviewed by:
- [x] @hixi


web_frontend/osmaxx/excerptexport/tests/views/test_views_with_dowload_links.py
```diff
@@ -0,0 +1,122 @@
...
+    user = django_user_model.objects.create_user(username='lauren', password='lauri', email=None)
+    client.login(username='lauren', password='lauri')
+    client.user = user
+    return client
+
+
+@pytest.fixture
+def authorized_client(authenticated_client):
+    authenticated_client.user.groups.add(Group.objects.get(name=FRONTEND_USER_GROUP))
+    return authenticated_client
```
das-g: I'd like to move above fixtures to a shared location (`conftest.py`?), but I think I'll need some help with that.

hixi:
:+1:

PR #405, opened 2016-02-19
Feature/#96 arbitrary polygon support

hixi:
This adds countries back to the frontend. You cannot cut out countries in the current state this is in, because it is still missing the multipolygon-json-backend-client that is needed to receive such a request.

Review is possibly easier from [#751db86](https://github.com/geometalab/osmaxx-frontend/commit/751db86e42b1684ee9ea773acc9081747899e16a) and [#4a5a0c5](https://github.com/geometalab/osmaxx-frontend/commit/4a5a0c5dd8e1a3119badd3ccab4742e0dbca5b12)

Reviewed by:
- [x] @das-g


web_frontend/osmaxx/countries/_settings.py
```diff
@@ -0,0 +1,9 @@
+import os
+
+from django.conf import settings
+
+POLYFILE_LOCATION = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'data', 'polyfiles')
+
+# only for testing, not documented
```
das-g: What does 'not documented' want to tell me here?

web_frontend/osmaxx/countries/models.py
```diff
@@ -0,0 +1,22 @@
...
+
+class Country(models.Model):
+    name = models.CharField(verbose_name=_('name'), max_length=100)
+    polyfile = InternalCountryFileField(verbose_name=_('polyfile'))
+    simplified_polygon = models.MultiPolygonField(verbose_name=_('simplified area'))
+
+    @property
+    def type_of_geometry(self):
+        """
+        Only needed for easier differentiation on the javascript side, has no real value otherwise.
```
das-g: 'discrimination' or more politically correct 'distinction' might be clearer terms than 'differentiation', which besides being a synonym for the former terms could also mean the computation of a mathematical function's derivative or becoming active in a wider field or more fields of business.

web_frontend/osmaxx/countries/utils.py
```diff
@@ -0,0 +1,75 @@
...
+        if simplify_tolerance:
+            poly = poly.simplify(tolerance=simplify_tolerance, preserve_topology=True)
+            # simplifying can lead to a polygon. Assure, it stays a multipolygon.
+            if poly and isinstance(poly, Polygon):
+                poly = MultiPolygon(poly)
+    return GEOSGeometry(poly)
+
+
+def parse_poly_string(poly_string):
+    return parse_poly(poly_string.split(os.linesep))
```
das-g: Should this be `os.linesep` or a fixed line separator as found in the Polygon Filter files we download?

Or should we split at anything that can be interpreter as a line separator (Python's "universal newlines" notion) with [`poly_string.splitlines()`](https://docs.python.org/3/library/stdtypes.html#str.splitlines)?

web_frontend/osmaxx/countries/_settings.py
hixi: That this is left out of the documentation (be it in the Readme or elsewhere) on purpose. But I agree, this can just be left out, it's nothing that helps to understand the code.

web_frontend/osmaxx/countries/models.py
hixi: Good point, didn't think of mathematics.

web_frontend/osmaxx/countries/utils.py
```diff
@@ -0,0 +1,75 @@
...
+from osmaxx.countries._settings import POLYFILE_LOCATION
+
+POLYFILE_ENDING = '.poly'
+
+
+def get_polyfile_name_to_file_mapping():
+    polyfile_mapping = {}
+    for possible_polyfile in os.listdir(POLYFILE_LOCATION):
+        if possible_polyfile.endswith(POLYFILE_ENDING):
+            name = possible_polyfile.split(POLYFILE_ENDING)[0]
```
das-g: Should we
```python
assert len(possible_polyfile.split(POLYFILE_ENDING)) == 1
```
?

web_frontend/osmaxx/tests/countries/views_test.py
```diff
@@ -0,0 +1,20 @@
+import pytest
+from rest_framework import status
+from rest_framework.reverse import reverse
+
+from osmaxx.countries.models import Country
+
+
+@pytest.mark.django_db
+def test_detail_country_contains_simplified_polygon(authenticated_client):
```
das-g: maybe `…_country_detail_…` instead of `…_detail_country_…`?

web_frontend/osmaxx/tests/countries/views_test.py
```diff
@@ -0,0 +1,20 @@
...
+    # Only Monaco will be available in countries for testing.
+    assert Country.objects.count() == 1
+
+    country = Country.objects.first()
+    url = reverse('excerptexport_api:country-detail', kwargs=dict(pk=country.id))
+    response = authenticated_client.get(url, format='json')
+    assert response.status_code == status.HTTP_200_OK
+    data = response.data
+    assert data.get('type') == "Feature"
+    assert data.get('id') is not None
```
das-g: Can't we assume a bit more than `is not None`? I'd expect it to be `== country.id`

das-g:
Should we also place a license file for the production `*.poly` files? (Did we have one in osmaxx-conversion-service? Under what terms did we obtain the boundary data?)

web_frontend/osmaxx/countries/utils.py
hixi: I asked myself the same question, but couldn't figure out which one is to be used. I opted for os.linesep because that would work on any system, but universal newlines is in fact more correct.

Didn't realize `string.splitlines()` existed, thanks for the input ;-)

hixi:
License file(s) would be a good idea, but AFAIK we had none so far, so we should probably track, where they came from in the first place (I think they were from `Wambacher`, but we definitively should check that).

web_frontend/osmaxx/countries/utils.py
hixi: Didn't think of that. But actually, this seems problematic, because the filename can be virtually any string. 
I think, `possible_polyfile.strip(POLYFILE_ENDING)` would make more sense, right?

web_frontend/osmaxx/tests/countries/views_test.py
hixi: :+1:

web_frontend/osmaxx/tests/countries/views_test.py
hixi: :+1:

hixi:
closing, already included in #475

PR #410, opened 2016-03-16
Feature/document local prerequisites

das-g:
#### Reviewed by:
- [x] @hixi
- [x] @wasabideveloper


docs/project-development-environment.md
```diff
@@ -2,7 +2,19 @@
 
 ## Local prerequisites
 
-Docker and docker-compose is required to be installed.
+| dependency | supported versions | Installation recommendation for Ubuntu 14.04 |
+| ---------- | ------------------ | ---------------------------------------------|
+| <a name='dependency_docker'>docker</a> | 1.10 (1.9 may still work, too) | Follow [the official Docker installation instructions for Linux](https://docs.docker.com/linux/step_one/) |
+| <a name='dependency_docker-compose'>docker-compose</a> | 1.6 or better | Install system-wide via <a href='#dependency_pip'>`pip`</a> (**not** via `pip3`!): <pre class="highlight highlight-source-shell">sudo pip install docker-compose</pre> |
```
wasabideveloper: "1.6 or newer" (let's hope the new versions are better ;-)

wasabideveloper: Why pip and not pip3? Please add the reason.

das-g: > "1.6 or newer" (let's hope the new versions are better ;-)

Actually, I consciously chose "or better" exactly because newer versions in the past have not always been better (for us / for our then-current docker-compose.yml files) due to (sometimes subtle) backwards incompatibilities. Stuffing this semantic into the phrasing is rather obscure, though, so I'll change this to just "1.6".

As the column is labeled "supported versions", not "required versions", that will capture the intended meaning sufficiently well. We can start "supporting" other versions of docker-compose once they're out and have been tested by us.

docs/project-development-environment.md
```diff
@@ -2,7 +2,19 @@
 
 ## Local prerequisites
 
-Docker and docker-compose is required to be installed.
+| dependency | supported versions | Installation recommendation for Ubuntu 14.04 |
+| ---------- | ------------------ | ---------------------------------------------|
+| <a name='dependency_docker'>docker</a> | 1.10 (1.9 may still work, too) | Follow [the official Docker installation instructions for Linux](https://docs.docker.com/linux/step_one/) |
```
wasabideveloper: To get more space you could put the version in the same field as the name but on the next row.

docs/project-development-environment.md
```diff
@@ -2,7 +2,19 @@
...
+| <a name='dependency_docker-compose'>docker-compose</a> | 1.6 | Install system-wide via <a href='#dependency_pip'>`pip`</a> (**not** via `pip3`! `docker-compose` is implemented in Python 2.): <pre class="highlight highlight-source-shell">sudo pip install docker-compose</pre> |
+| <a name='dependency_pg_config'>pg_config</a> (required by psycopg2) |  | <pre class="highlight highlight-source-shell">sudo apt install libpq-dev</pre> |
```
wasabideveloper: Maybe it makes sense to add a field for what the dependency is used or to split the table and make titles.

docs/project-development-environment.md
```diff
@@ -2,7 +2,19 @@
...
+| <a name='dependency_geos_c'>GEOS C library</a> | whatever [GeoDjango supports](https://docs.djangoproject.com/en/1.9/ref/contrib/gis/install/geolibs/#installing-geospatial-libraries) | <pre class="highlight highlight-source-shell">sudo apt install python3-gdal</pre> will pull this and other required libraries in. `python3-gdal` itself is not required. Thus, if you prefer a more minimal installation, only install `libgeos-c1`. |
+| <a name='dependency_gdal'>GDAL library</a> | whatever [GeoDjango supports](https://docs.djangoproject.com/en/1.9/ref/contrib/gis/install/geolibs/#installing-geospatial-libraries) | <pre class="highlight highlight-source-shell">sudo apt install python3-gdal</pre> will pull this and other required libraries in. `python3-gdal` itself is not required. Thus, if you prefer a more minimal installation, only install `libgdal1h`. |
+| <a name='dependency_pip'>pip</a> (Python 2) |  | <pre class="highlight highlight-source-shell">sudo apt install python-pip</pre> |
```
wasabideveloper: Maybe it makes sense to list the dependencies in the order it should be installed. Pip should be installed first, not?

docs/project-development-environment.md
```diff
@@ -2,7 +2,19 @@
...
 
 For committing and using the pre-commit hook (which really should be used) flake8 needs to be installed on
```
wasabideveloper: It makes sense to create the same table as above for these dependencies.

wasabideveloper:
Maybe it makes sense to rework the whole document `project-development-environment.md`. There is another chapter about docker 1.5 and I'm not shure it's outdated.

docs/project-development-environment.md
das-g: I decided to line-wrap commands (2b6566926c52f1ce5642b621ed0de773a09504f8) instead of denormalizing the table. One that does not suffice anymore, I'd extract the installation recommendations to sections outside the table and link to those from the table.

docs/project-development-environment.md
das-g: I, too, have considered that, in two variants:

### listing reverse dependencies (i.e. the dependees), too
 I haven't yet figured out how to do that without making the table too verbose and convoluted. (Some elements are dependencies of several others. For some elements, the choice of how to install them affects what other elements need to be installed first.)

### coarsely categorize dependencies
e.g.
- **install-time dependency**, e.g. pip and &mdash;less obviously&mdash; pg_config
- **build-time dependency**, e.g. docker
- **local (test suit) runtime dependency**, e.g. pytest (would also include general runtime dependencies that have to be installed system-wide for running the test suit, like GDAL)

Many elements would belong two more than one of these categories, so separate tables would mean listing them multiple times.

docs/project-development-environment.md
das-g: For some elements, the choice of how to install them affects what other elements need to be installed first. Also depending on how you install the former, the latter might be pulled in automatically by the installation system or not. Thus, we would have to list some optional elements before some mandatory elements with the decision whether you want an optional one or not depends on how you want to install a later listed mandatory one. That would be rather confusing, wouldn’t it?

Therefore, I tried to list them in the order you have to make your decisions. Not yet sure whether that makes more sense, though.

docs/project-development-environment.md
das-g: Probably, but that's something I might do later. For this PR, I declare it out-of-scope.

das-g:
> Maybe it makes sense to rework the whole document `project-development-environment.md`. There is another chapter about docker 1.5 and I'm not sure it's outdated.

Yeah. @hixi and I will have to comb through it. Not as part of this PR, though.

PR #411, opened 2016-03-17
Feature/docker compose env files in subdir

das-g:
### Reviewed by:
- [x] @hixi


Makefile
```diff
@@ -75,3 +99,4 @@ down-pg_translit:
 clean: down-redis down-pg_translit down-pg
 	find . -name __pycache__ -exec rm -rf {} +
 	find . -name "*.pyc" -exec rm -rf {} +
+	rm compose-env/*.env
```
hixi: This isn't what I would like: if I have defined own settings (ie. testing the production settings locally) then they will be removed.

das-g: You're right. I didn't consider that. Even worse for production. (do we back up our `*.env` files there?)

I've removed that line from the recipe: 67377736f9580dfc8071608e805d9e2862023a43

PR #412, opened 2016-03-23
Feature/raise_for_status in API_client

das-g:
Reviewed by:
- [x] @hixi

### DIFF
--- a/osmaxx/api_client/API_client.py
+++ b/osmaxx/api_client/API_client.py
@@ -11,11 +11,8 @@ class RESTApiJWTClient:
 
     Can be either extended through inheritance or created using the params `service_base` and `login_url`
 
-    returns the response object and errors dictionary: if errors are None, all is fine.
-
     :param service_base: the base url
     :param login_url: the relative path to the login url
-    :returns response object
     """
 
     service_base = 'http://localhost/api/'
--- a/osmaxx/api_client/API_client.py
+++ b/osmaxx/api_client/API_client.py
@@ -31,31 +31,29 @@ class RESTApiJWTClient:
         }
         self.client = requests.session()
         self.token = None
-        self.errors = None
 
     def get(self, url, params=None, **kwargs):
         response = requests.get(self._to_fully_qualified_url(url), params=params, **self._data_dict(**kwargs))
-        self.errors = get_error(response)
+        raise_for_status(response)
         return response
 
     def post(self, url, json_data=None, **kwargs):
         response = requests.post(self._to_fully_qualified_url(url), json=json_data, **self._data_dict(**kwargs))
-        self.errors = get_error(response)
+        raise_for_status(response)
         return response
 
     def options(self, url, **kwargs):
-        response = get_error(
+        response = raise_for_status(
             requests.options(self._to_fully_qualified_url(url), **self._data_dict(**kwargs))
         )
-        self.errors = get_error(response)
+        raise_for_status(response)
         return response
 
     def auth(self, username, password):
         login_url = self._to_fully_qualified_url(self.login_url)
         login_data = dict(username=username, password=password, next=self.service_base)
         response = self.post(login_url, json_data=login_data, headers=dict(Referer=login_url))
-        if not self.errors:
-            self.token = response.json().get('token')
+        self.token = response.json().get('token')
         return response
 
     def authorized_get(self, url, params=None, **kwargs):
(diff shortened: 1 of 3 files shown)


osmaxx/api_client/API_client.py
```diff
@@ -4,6 +4,8 @@
 
 logger = logging.getLogger(__name__)
 
+HTTPError = requests.HTTPError
```
hixi: why not import this directly?
`from requests import HTTPError`

osmaxx/api_client/API_client.py
```diff
@@ -112,20 +77,6 @@ def _make_request_authorized(self):
...
-        if self._request_successful(response):
-            try:
-                json_response = response.json()
-                return json_response.get(key, None)
-            except ValueError:
-                logging.error('API job creation failed because of an JSON decoding Issue.', str(response.__dict__))
-                return None
-        return None
+
+def errors(http_error):
```
hixi: :+1:

osmaxx/api_client/API_client.py
das-g: Because that would lead to
> Running flake8 code linting
> osmaxx/api_client/API_client.py:4:1: F401 'HTTPError' imported but unused
> flake8 failed

osmaxx/api_client/API_client.py
das-g: I wonder whether a different name would make the usage more comprehensible. `reasons_for`, maybe?

osmaxx/api_client/conversion_api_client.py
```diff
@@ -85,19 +76,20 @@ def create_job(self, extraction_order, request):
             }
         })
         self.login()
-        response = self.authorized_post(self.conversion_job_url, json_data=request_data)
-        if self.errors:
-            logging.error('API job creation failed.', response)
+        try:
```
hixi: Even though this doesn't thematically belong to this pull request, this entire code block doesn't work with the new API anymore.

osmaxx/api_client/API_client.py
hixi: And flake is right, it isn't used. Why not import it then at the place it actually _is_ used?

osmaxx/api_client/API_client.py
hixi: `reasons_for` would be adequate as well, though `errors(http_error)` seems readable enough for me - I'll leave that naming to your liking ;-) .

osmaxx/api_client/API_client.py
das-g: Because the place(s) where it is used shouldn't have to care that `API_client.HTTPError` is an alias and actually just a `requests.HTTPError`.

The aim is decoupling: `API_client.py` should be the only file dependent on `requests`. Thus if `requests` changes or if we replace `requests`, it should be possible to 'hide' that change from the rest of our system by only changing `API_client.py` accordingly.

osmaxx/api_client/conversion_api_client.py
das-g: I'm aware of that and plan to tackle that in future commits on branch `feature/conversion-api-client_reimplementation`.

osmaxx/api_client/API_client.py
```diff
@@ -66,31 +56,6 @@ def authorized_post(self, url, json_data=None, **kwargs):
...
-
-    def _get_error(self, response):
-        try:
-            response.raise_for_status()
-            error = None
-        except requests.HTTPError as e:
-            try:
-                error = e.response.json()
-            except:
-                error = {'unknown': 'Unknown error happened. Please try again.'}
```
das-g: Should the `reasons_for(http_error)` implementation be guarded like this, too?

osmaxx/api_client/API_client.py
hixi: That intention would be much clearer to me, if we defined our own class in `API_client.py` and would document this as eg. `thin wrapper` around the requests HTTPError. Like it stands currently, I would assume I can actually remove the declaration as it isn't used.

hixi: Thinking a bit more about this:
This seems for me the wrong encapsulation; we don't control what is being raised, we just assume that `raise_for_status` is actually raising a `HTTPError`. This is very implicit, so we either should re-raise it with our own Exception-Class (or wrapper thereof) or just drop the whole game: our class raises another class' exception, that is nothing uncommon. We are relying very heavily on the requests library, no need to hide that away.

osmaxx/api_client/API_client.py
hixi: No, let's not do that again; this was only needed because we had to have an explicit response. Now it should just fail loudly IMO.

osmaxx/api_client/conversion_api_client.py
hixi: :+1:

osmaxx/api_client/API_client.py
das-g: importing from requests where used with d7548af

osmaxx/api_client/API_client.py
das-g: renamed with e003093

hixi: :+1:

Issue #343, opened 2015-12-23
Make API client DRY


wasabideveloper:
extract duplicated code of `get`, `post` and `delete` methods.

PR #415, opened 2016-03-26
Feature/#343 remove duplication in rest api client http verb methods

das-g:
Not sure whether 92f417d really is an improvement. (Maybe there was more severe duplication back when #343 was filed?) But I guess the tests are worth keeping, so if we don't want 92f417d, I'd suggest to still [merge branch `add_rest_api_client_unit_tests`](https://github.com/geometalab/osmaxx-frontend/compare/combine_repos...add_rest_api_client_unit_tests?expand=1).

#### Reviewed by:
- [x] @hixi
- [ ] @beneditatan

### DIFF
--- a/osmaxx/api_client/API_client.py
+++ b/osmaxx/api_client/API_client.py
@@ -30,12 +30,15 @@ class RESTApiJWTClient:
         self.token = None
 
     def get(self, url, params=None, **kwargs):
-        response = requests.get(self._to_fully_qualified_url(url), params=params, **self._data_dict(**kwargs))
-        response.raise_for_status()
-        return response
+        return self._request(requests.get, url, dict(params=params), kwargs)
 
     def post(self, url, json_data=None, **kwargs):
-        response = requests.post(self._to_fully_qualified_url(url), json=json_data, **self._data_dict(**kwargs))
+        return self._request(requests.post, url, dict(json=json_data), kwargs)
+
+    def _request(self, method, url, payload, kwargs):
+        kwargs = self._data_dict(**kwargs)
+        kwargs.update(payload)
+        response = method(self._to_fully_qualified_url(url), **kwargs)
         response.raise_for_status()
         return response
 
--- a/osmaxx/api_client/API_client.py
+++ b/osmaxx/api_client/API_client.py
@@ -5,7 +5,7 @@ import requests
 logger = logging.getLogger(__name__)
 
 
-class RESTApiJWTClient:
+class RESTApiClient:
     """
     REST Client Base Class With JWT enabled authentication
 
@@ -27,7 +27,6 @@ class RESTApiJWTClient:
             'Content-Type': 'application/json; charset=UTF-8',
         }
         self.client = requests.session()
-        self.token = None
 
     def get(self, url, params=None, **kwargs):
         return self._request(requests.get, url, dict(params=params), kwargs)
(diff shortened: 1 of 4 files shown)


tests/api_client/rest_api_client_test.py
```diff
@@ -0,0 +1,74 @@
...
+    c = RESTApiJWTClient('http://example.com/service/uri_base/')
+    response = c.get('get/example')
+    assert response.request.headers['Content-Type'] == 'application/json; charset=UTF-8'
+
+
+def test_get_sends_payload_in_query_string(requests_mock):
+    requests_mock.get(ANY)
+    c = RESTApiJWTClient('http://example.com/service/uri_base/')
+    response = c.get('get/example', params={'some_key': b'a value'})
+    assert response.request.url == 'http://example.com/service/uri_base/get/example?some_key=a+value'
```
das-g: @hixi Until now, we don't actually use this capability. Should the explicit code for it be removed (it would continue to work because we forward `**kwargs`) or should it be retained, so the API client has a somewhat complete interface?

tests/api_client/rest_api_client_test.py
```diff
@@ -0,0 +1,74 @@
...
+    c = RESTApiClient('http://example.com/service/uri_base/')
+    response = c.get('get/example')
+    assert response.request.headers['Content-Type'] == 'application/json; charset=UTF-8'
+
+
+def test_get_sends_payload_in_query_string(requests_mock):
+    requests_mock.get(ANY)
+    c = RESTApiClient('http://example.com/service/uri_base/')
+    response = c.get('get/example', params={'some_key': b'a value'})
+    assert response.request.url == 'http://example.com/service/uri_base/get/example?some_key=a+value'
```
das-g: @hixi Until now, we don't actually use this capability. Should the explicit code for it be removed (it would continue to work because we forward `**kwargs`) or should it be retained, so the API client has a somewhat complete interface?

hixi: In this case I'm in favor of keeping it - we're very likely to actually be using this (or at least we should in the end). Even if not, having a more complete API is a win in my opinion.

osmaxx/api_client/API_client.py
```diff
@@ -27,33 +27,20 @@ def __init__(self, service_base=None, login_url=None):
...
-        response.raise_for_status()
-        return response
+        return self._request(requests.post, url, dict(json=json_data), kwargs)
 
-    def auth(self, username, password):
-        login_url = self._to_fully_qualified_url(self.login_url)
-        login_data = dict(username=username, password=password, next=self.service_base)
-        response = self.post(login_url, json_data=login_data, headers=dict(Referer=login_url))
-        self.token = response.json().get('token')
+    def _request(self, method, url, payload, kwargs):
```
hixi: Even though this is a bit more "obscure" I like the passing in functions to keep the code (more) dry. :+1:

tests/api_client/rest_api_client_test.py
hixi: In this case I'm in favor of keeping it - we're very likely to actually be using this (or at least we should in the end). Even if not, having a more complete API is a win in my opinion.

tests/conftest.py
```diff
@@ -169,6 +169,13 @@ def pytest_configure():
 
 # if any global fixtures are needed, add them below
 
+@pytest.yield_fixture
+def requests_mock():
+    import requests_mock
+    with requests_mock.mock() as m:
+        yield m
```
hixi: :+1:

osmaxx/api_client/API_client.py
```diff
@@ -27,33 +27,20 @@ def __init__(self, service_base=None, login_url=None):
...
+        kwargs = self._data_dict(**kwargs)
+        kwargs.update(payload)
+        response = method(self._to_fully_qualified_url(url), **kwargs)
```
das-g: In Python &ge;3.5, the above 3 lines could be boiled down to a single one:

            response = method(self._to_fully_qualified_url(url), **payload, **kwargs)
(Other than the current implementation, this wouldn't allow the same keys to be present in both dicts.)
See [PEP 448](https://www.python.org/dev/peps/pep-0448/#rationale) Additional Unpacking Generalizations.

Issue #133, opened 2015-07-13
maybe: use wheel (pip7+) to release a wheel tar/zip with dependencies included


ghost:
See: https://lincolnloop.com/blog/fast-immutable-python-deployments/

https://pip.pypa.io/en/stable/reference/pip_wheel.html#examples

njordan-hsr:
First step would be to add a `setup.py` for the project.

hixi:
They way we are working currently, I think this is not going to happen - and if it does, we don't need an issue for it.

PR #427, opened 2016-04-05
Feature/#133 use the world db instead of pbfs directly

hixi:
Add support for using the world database instead of the the pbf directly.

Reviewed by:
- [x] @das-g


docker-compose.yml
```diff
@@ -1,85 +1,163 @@
...
+  osm_data: {}
+  osm_cache: {}
+  database-postgis-data: {}
+networks:
+  osmaxx-internal:
+    driver: bridge
+  osm-world:
+    driver: bridge
+  osm-world-internal:
+    driver: bridge
```
das-g: This network layout separates some containers from each other, but still let's containers access each other that would not have to. One could consider splitting the `osmaxx-internal` network into 5 separate networks:
- network `osmaxx-frontend-internal` between containers
  - `frontend`
  - `frontenddatabase`
- network `osmaxx-mediator-internal` between containers
  - `mediator`
  - `mediatordatabase`
- network `osmaxx-conversion-rest-api` (or `osmaxx-core`? name T.B.D.) between containers
  - `frontend`
  - `mediator`
- network `osmaxx-conversionqueue-push` (or a better name T.B.D.) between containers
  - `mediator`
  - `conversionserviceredis`
- network `osmaxx-conversionqueue-pull` (or a better name T.B.D.) between containers
  - `worker`
  - `conversionserviceredis`

I've discussed that with @hixi and we came to the conclusions:
- Accidental unwarranted access between containers is sufficiently deterred by the need to link containers first. (One can deliberately work around that, but not unintentionally.)
- Neither network layout adds much to the application security, as the sensitive data is held in the front end and its data base.
- System security isn't inherently increased by the layout with more networks: If someone manages to break out of docker, they can probably do that from the front end already.

So for now, we'll stick with the network layout introduced in this PR.

osmaxx/conversion/converters/converter_gis/bootstrap/bootstrap.py
```diff
@@ -1,30 +1,28 @@
...
+from osmaxx.conversion.converters.converter_gis.helper.osm_importer import OSMImporter
+from osmaxx.utils import changed_dir, polyfile_helpers
 
 
-def boostrap(pbf_file_path):
+def boostrap(area_polyfile_string):
     with changed_dir(os.path.dirname(__file__)):
-        bootstrapper = BootStrapper(pbf_file_path=pbf_file_path)
+        bootstrapper = BootStrapper(area_polyfile_string)
         bootstrapper.bootstrap()
```
das-g: Could the `with changed_dir(...)` call be moved into `BootStrapper.bootstrap(...)`? I don't think the `BootStrapper` initialization is sensitive to the current working dir.

osmaxx/conversion/converters/converter_gis/bootstrap/sql/functions/0040_interpolate_addresses.sql
```diff
@@ -7,7 +7,7 @@ drop table if exists addr_interpolated;
 create table addr_interpolated as
 -- BEGIN
 with addr_interpolation_line as (
-  select osm_id,
+    select osm_id,
```
das-g: Please keep the `select` aligned with the `from` and `where` lines below.

osmaxx/conversion/converters/converter_gis/gis.py
```diff
@@ -32,22 +31,19 @@ def __init__(self, *, conversion_format, out_zip_file_path, base_file_name, out_
         self._static_directory = os.path.join(os.path.abspath(os.path.dirname(__file__)), 'static')
 
     def create_gis_export(self):
-        with pbf_area(osmosis_polygon_file_content=self._polyfile_string) as pbf_path:
```
das-g: This was the only production code using the `pbf_area` CM. The test code seems to access it to test the CM, not as a tool for testing other stuff. So the CM implementation could now be removed as well.

osmaxx/conversion/converters/converter_gis/helper/osm_importer.py
```diff
@@ -0,0 +1,94 @@
+from itertools import chain
+from sqlalchemy import MetaData, Table, create_engine, func
+from sqlalchemy.engine.url import URL
+from sqlalchemy.sql import select, insert
+# import needed for reflection to work correctly with spatial types
+from geoalchemy2 import Geometry, Geography  # noqa
```
das-g: Consider asserting for these classes near where they are used for reflection. Like that, they become used and we don't need the `# noqa`. And if the code that needs them ever gets removed, we'd notice that the imports are actually unused again.

osmaxx/conversion/converters/converter_gis/helper/osm_importer.py
```diff
@@ -0,0 +1,94 @@
...
+        self._osm_derived_tables = ['osm_point', 'osm_line', 'osm_polygon', 'osm_roads']
+        self._osm_base_tables = ['osm_ways', 'osm_nodes']
+
+        _world_db_connection_parameters = {
+            'username': 'gis',
+            'password': 'gis',
+            'port': 5432,
+            'database': 'gis',
+            'host': 'world-database',
+        }
```
das-g: Might be slightly more readable with the `dict(...)` constructor.

osmaxx/conversion/converters/converter_gis/helper/osm_importer.py
```diff
@@ -0,0 +1,94 @@
...
+        world_db_connection = URL('postgresql', **_world_db_connection_parameters)
+        self._world_db_engine = create_engine(world_db_connection, use_native_hstore=True)
+        self._world_db_meta_data = MetaData()
+
+        _local_db_connection_parameters = {
+            'username': 'postgres',
+            'password': 'postgres',
+            'port': 5432,
+            'database': 'osmaxx_db',
+        }
```
das-g: Might be slightly more readable with the `dict(...)` constructor.

osmaxx/conversion/converters/converter_gis/bootstrap/bootstrap.py
hixi: good catch :-)

osmaxx/conversion/converters/converter_gis/helper/osm_importer.py
```diff
@@ -0,0 +1,94 @@
...
+        local_db_connection = URL('postgresql', **_local_db_connection_parameters)
+        self._local_db_engine = create_engine(local_db_connection, use_native_hstore=True)
+        self._table_metas = {}
+
+        for to_be_created_table in self._osm_derived_tables + self._osm_base_tables:
```
das-g: Consider using a dictionary comprehension to combine assignment and content generation.

osmaxx/conversion/converters/converter_gis/helper/osm_importer.py
```diff
@@ -0,0 +1,94 @@
...
+            self._table_metas[to_be_created_table] = Table(
+                to_be_created_table,
+                self._world_db_meta_data,
+                autoload=True,
+                autoload_with=self._world_db_engine,
+                postgresql_with_oids=True
```
das-g: Is `WITH OIDS` still needed? If so, please add a short comment that states for what. (HSTORE reflection, if I understood you correctly?)

osmaxx/conversion/converters/converter_gis/gis.py
hixi: didn't think of that, getting rid of more code :-)

osmaxx/conversion/converters/converter_gis/helper/osm_importer.py
```diff
@@ -0,0 +1,94 @@
...
+            )
+
+    def load_area_specific_data(self, *, extent):
```
das-g: :+1: for abstract-above-specific method order in source file :smiley:

osmaxx/conversion/converters/converter_gis/helper/osm_importer.py
hixi: Just making sure, you would prefer
```python
self._table_metas = {
    to_be_created_table: Table(
                to_be_created_table,
                self._world_db_meta_data,
                autoload=True,
                autoload_with=self._world_db_engine,
                postgresql_with_oids=True
}
for to_be_created_table in self._osm_derived_tables + self._osm_base_tables
]
```

over the current for loop code?

osmaxx/conversion/converters/converter_gis/helper/osm_importer.py
hixi: agreed, changing.

osmaxx/conversion/converters/converter_gis/helper/osm_importer.py
hixi: agreed, changing.

osmaxx/conversion/converters/converter_gis/helper/osm_importer.py
hixi: yes, this was the case. Since lot of code change since those initial steps, I'm checking if this really is needed.

hixi: Not needed anymore, auto-discovery works, removing. Thanks for the catch :-)

osmaxx/conversion/converters/converter_gis/helper/osm_importer.py
hixi: good idea.

osmaxx/conversion/converters/converter_gis/helper/osm_importer.py
```diff
@@ -0,0 +1,94 @@
...
+        self._load_base_tables()
+
+    def _create_tables_on_local_db(self):
+        self._world_db_meta_data.create_all(self._local_db_engine)
+
+    def _load_tables(self, extent):
+        for table_name in self._osm_derived_tables:
+            table_meta = self._table_metas[table_name]
+            query = select([table_meta])
+            query = query.where(func.ST_Intersects(table_meta.c.way, extent.ewkt))
```
das-g: Is `.c.way` correct for table `osm_point`, too?

Maybe the [`GEOMETRY_COLUMNS` VIEW](http://postgis.net/docs/manual-2.0/using_postgis_dbmanagement.html#geometry_columns) could be used to identify the respective geography column dynamically.

hixi: I believe so:
```sql
-- Column: way
-- ALTER TABLE public.osm_point DROP COLUMN way;
ALTER TABLE public.osm_point ADD COLUMN way geometry(Point,4326);
```

osmaxx/conversion/converters/converter_gis/helper/osm_importer.py
das-g: Well, without the syntax errors
```python
        self._table_metas = {
            to_be_created_table: Table(
                to_be_created_table,
                self._world_db_meta_data,
                autoload=True,
                autoload_with=self._world_db_engine,
                postgresql_with_oids=True,
            )
            for to_be_created_table in self._osm_derived_tables + self._osm_base_tables   
        }
```

or more comprehensible
```python
    def __init__(self):
        # ....
        self._table_metas = {
            to_be_created_table: self._world_db_table_definition(table_name=to_be_created_table)
            for to_be_created_table in self._osm_derived_tables + self._osm_base_tables   
        }

    # ...
    
    def _world_db_table_definition(self, table_name):
        return Table(
            table_name,
            self._world_db_meta_data,
            autoload=True,
            autoload_with=self._world_db_engine,
            postgresql_with_oids=True,
        )
```

osmaxx/conversion/converters/converter_gis/helper/osm_importer.py
```diff
@@ -0,0 +1,94 @@
...
+            self._execute_and_insert_into_local_db(query, table_meta)
+
+    def _load_base_tables(self):
+        osm_line = self._table_metas['osm_line']
+        osm_ways = self._table_metas['osm_ways']
+        osm_nodes = self._table_metas['osm_nodes']
+
+        local_line_osm_ids_query = self._local_db_engine.execute(
+            select([osm_line.c.osm_id])
+        )
```
das-g: `local_line_osm_ids_query` is not a query

osmaxx/conversion/converters/converter_gis/helper/osm_importer.py
```diff
@@ -0,0 +1,94 @@
...
+                    node_ids = chain.from_iterable(array_rows)
+                    osm_nodes_query = select([osm_nodes]).where(
+                        osm_nodes.c.id.in_(node_ids)
+                    )
+                    self._execute_and_insert_into_local_db(osm_nodes_query, osm_nodes)
+
+    def _execute_and_insert_into_local_db(self, query, table_meta):
+        query_result = self._world_db_engine.execute(query)
+        row_count = query_result.rowcount
+        if row_count > 0:
```
das-g: Inline local variable `row_count`. It doesn't add to the readability, as the property is already adequately named.

osmaxx/conversion/viewsets.py
```diff
@@ -9,7 +9,7 @@
 
 
 class JobViewSet(viewsets.ModelViewSet):
-    queryset = Job.objects.all()
+    queryset = Job.objects.all().order_by('-id')
```
das-g: This means newest first, I guess?

osmaxx/utils/polyfile_helpers.py
```diff
@@ -1,6 +1,24 @@
 import os
+from django.contrib.gis.geos import Polygon, MultiPolygon, GEOSGeometry
 
-from django.contrib.gis.geos import MultiPolygon, Polygon
+from osmaxx.countries._settings import POLYFILE_LOCATION
+
+POLYFILE_ENDING = '.poly'
```
das-g: Maybe name this `POLYFILE_FILENAME_EXTENSION` to distinguish it from the ending of polygon filter file format file contents.

osmaxx/conversion/viewsets.py
hixi: Yes.

osmaxx/utils/polyfile_helpers.py
hixi: :+1:

tests/clipping_area/to_polyfile_test.py
```diff
@@ -3,8 +3,8 @@
 
 from django.contrib.gis.geos import MultiPolygon, Polygon, MultiLineString, LineString
 
+import osmaxx.utils.polyfile_helpers
 from osmaxx.clipping_area.to_polyfile import create_poly_file_string
-from osmaxx.clipping_area import utils
```
das-g: Why not
```python
from osmaxx.utils.polyfile_helpers import parse_poly_string
```
? When using `parse_poly_string` below, we don't need the additional context of where it's from to tell what its name means.

osmaxx/conversion/converters/converter_gis/helper/osm_importer.py
hixi: This doesn't work, since it is being returned at the end of the function.

das-g:
Is #133 done with this PR or is more necessary?

hixi:
This should actually already be all :-)

PR #429, opened 2016-04-06
re-record VCR casettes

das-g:
re-record VCR casettes against mediator and worker of osmaxx-conversion-service revision 1dd8fae6b1fa76b573b635bb18d1d48b558837b3 (i.e. pre API change of #406)

### Reviewed by
- [ ] @hixi


das-g:
Actually, let's use this as a dead-end branch for now instead of merging it.

PR #417, opened 2016-04-01
Feature/#406 conversion API client reimplementation: create_parametrization()

das-g:
connected to #406 (implements part of it)

#### Implements
* `ConversionApiClient.create_parametrization(...)`

#### Already implemented in #416
* `ExtractionOrder.forward_to_conversion_service(...)`
* `ConversionApiClient.create_boundary(...)`

#### Still to be implemented
(already being used by implementation of `ExtractionOrder.forward_to_conversion_service(...)`):
* `ConversionApiClient.api_client.create_job(...)`

### Reviewed by
- [x] @hixi
- [ ] @beneditatan

PR #416, opened 2016-03-30
Feature/#406 conversion API client reimplementation: create_boundary

das-g:
connected to #406 (implements part of it)

#### Implements
* `ExtractionOrder.forward_to_conversion_service(...)`
* `ConversionApiClient.create_boundary(...)`

#### Still to be implemented
(already being used by implementation of `ExtractionOrder.forward_to_conversion_service(...)`):
* `ConversionApiClient.create_parametrization(...)`
* `ConversionApiClient.api_client.create_job(...)`

### Reviewed by
- [x] @hixi
- [ ] @beneditatan

PR #430, opened 2016-04-07
Feature/#406 conversion api client reimplementation

das-g:
connected to #406 (implements part of it)

#### Implements
* `ConversionApiClient.api_client.create_job(...)`

#### Already implemented in #416 & #417 
* `ExtractionOrder.forward_to_conversion_service(...)`
* `ConversionApiClient.create_boundary(...)`
* `ConversionApiClient.create_parametrization(...)`

#### Still to be adapted
* `ExtractionOrder.forward_to_conversion_service(...)` (See FIXME comment in code)
    * This will require db schema changes (model change causing migrations), so that status can be tracked per (now format-specific) job instead of per extraction order.

### Reviewed by
- [x] @hixi


osmaxx/api_client/conversion_api_client.py
```diff
@@ -62,7 +78,8 @@ def _extraction_processing_overdue(progress, extraction_order):
         timeout_reached = timezone.now() > extraction_order.process_due_time
         return process_unfinished and timeout_reached
 
-    def create_job(self, extraction_order, request):
+    # TODO: replace this by a call to extraction_order.forward_to_conversion_service(request)
```
hixi: :+1:

Issue #438, opened 2016-04-15
let front-end properly react to conversion job status updates from revamped conversion service


das-g:
- [x] define what job status changes should cause what actions in the front end
  - Tracking job stati in `Export` model instances for later display (upon user request)
    - [ ] ~~identify possible states and allowed transitions~~
  - Notification of orderer by on-screen messages
    - probably for all or most status changes of individual exports / jobs
  - Notification of orderer by email
    - probably if the overall extraction order enters specific stati (computed from its exports' stati)
      - [x] Which ones? Only the final ones (finished or failed) as well as timed out?
      - [ ] ~~How to compute the overall status?~~
- [x] find out what status update notification the conversion service now sends to the callback URL
  - [x] If some of those needed for the actions found above aren't sent by the service now, file another issue to make the service send them.

das-g:
> - [x] define what job status changes should cause what actions in the front end
>   - Tracking job stati in `Export` model instances for later display (upon user request)
>     - [ ] ~~identify possible states and allowed transitions~~

We'll exactly mirror the status as sent by callback or determined by querying conversion service job resource (the latter is done if we detect stale statuses on the frontend).

> - [x] define what job status changes should cause what actions in the front end<br />[...]
>   - Notification of orderer by on-screen messages
>     - probably for all or most status changes of individual exports / jobs

Yes, for **all that the frontend gets notified about**. (The statuses that it _gets_ notified about might currently be only few, but that might change service-side. Not as part of this issue, though, but the frontend logic must already be able to handle that.)

> - [x] define what job status changes should cause what actions in the front end<br />[...]
>   - Notification of orderer by email
>     - probably if the overall extraction order enters specific stati (computed from its exports' stati)
>       - [x] Which ones? Only the final ones (finished or failed) as well as timed out?
>       - [ ] ~~How to compute the overall status?~~

Send email (with download links, as we did previously) when all exports of an extraction order are in a final state (finished, failed or timed out)

> - [x] find out what status update notification the conversion service now sends to the callback URL
>   - [x] If some of those needed for the actions found above aren't sent by the service now, file another issue to make the service send them.

We get a callback invocation at least for FINISHED. We won't care about the rest for now.

PR #439, opened 2016-04-15
Feature/#406 conversion api client reimplementation

### LINKED ISSUE #406: Adapt service calls in web frontend to match new API of conversion service

das-g:
closes #406

#### Implements
* call `create_job` with callback URL

#### Already implemented in #416, #417 & #430
* `ExtractionOrder.forward_to_conversion_service(...)`
* `ConversionApiClient.create_boundary(...)`
* `ConversionApiClient.create_parametrization(...)`
* `ConversionApiClient.api_client.create_job(...)`

#### Still to be adapted / fixed / re-implemented
* #438 properly react (again) to conversion job status updates

(I've filed that as a new issue, as it isn't part of what the previous conversion API client was.)

### Reviewed by
- [x] @hixi


conftest.py
```diff
@@ -1,3 +1,20 @@
+import pytest
+
+
 def pytest_addoption(parser):
     parser.addoption("--runslow", action="store_true",
                      help="run slow tests")
+    parser.addoption("--transactional", action="store_true",
```
hixi: :+1:

osmaxx/excerptexport/migrations/0021_export_file_format_verbose_name.py
```diff
@@ -0,0 +1,20 @@
...
+
+    dependencies = [
+        ('excerptexport', '0020_export_conversion_service_job_id'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='export',
+            name='file_format',
+            field=models.TextField(choices=[('fgdb', 'ESRI File Geodatabase'), ('shapefile', 'ESRI Shapefile'), ('gpkg', 'GeoPackage'), ('spatialite', 'SpatiaLite'), ('garmin', 'Garmin navigation & map data')], verbose_name='file format / data format'),
```
hixi: Why is this migration separately? Would it make sense join this change into the `0018_export.py`?

osmaxx/excerptexport/models/export.py
```diff
@@ -0,0 +1,41 @@
+from django.db import models
+from django.utils.translation import ugettext_lazy as _
+from rest_framework.reverse import reverse
+
+from osmaxx.conversion_api.formats import FORMAT_CHOICES
+
+
+class Export(models.Model):
+    """
+    Frontend-side representation of both, a export procedure in progress (or concluded) *and* the result of exporting.
```
hixi: :+1: I like the explanation!

osmaxx/job_progress/views.py
```diff
@@ -1,66 +1,13 @@
...
-            '',
-            'View the order at {order_url}',
-        ])).format(**substitutions)
-
-        emissary.error(message)
-        emissary.inform_mail(subject=finished_email_subject, mail_body=finished_email_body)
-    else:
-        message = _('Extraction order #{order_id} "{excerpt_name}" is now {order_state}.').format(**substitutions)
-        emissary.info(message)
+    # emissary = Emissary(recipient=export.extraction_order.orderer)
```
hixi: Does this code still serve a purpose later?

osmaxx/utils/frozendict.py
```diff
@@ -0,0 +1,12 @@
+from types import MappingProxyType as dictproxy  # noqa
```
hixi: is the noqa for lower_casing a ClassName?

tests/conftest.py
```diff
@@ -74,6 +75,7 @@ def pytest_configure():
             'rest_framework_gis',
             'rest_framework.authtoken',
             'crispy_forms',
+            'stored_messages',
```
hixi: why do you need `stored_message` in tests?

tests/excerptexport/migrations/conftest.py
```diff
@@ -0,0 +1,25 @@
...
+        return apps_at_state
+
+    yield _migrate
+
+    # Back to latest state for following tests:
+    call_command('migrate')
+
+
+@pytest.fixture
+def _executor():
```
hixi: This only seems to be used inside the `migrate` function above, what is the drawback when in-lining?

tests/excerptexport/migrations/test_0019_one_export_per_file_format.py
```diff
@@ -0,0 +1,46 @@
...
+    extraction_order = ExtractionOrder.objects.create(orderer_id=user.id)
+    extraction_order._extraction_configuration = json.dumps(extraction_configuration_with_formats)
+    extraction_order.save()
+    extraction_order_id = extraction_order.id
+
+    new_apps = migrate('excerptexport', '0019_one_export_per_file_format')
+    ExtractionOrder = new_apps.get_model('excerptexport', 'ExtractionOrder')  # noqa
+    extraction_order_after_migration = ExtractionOrder.objects.get(id=extraction_order_id)
+    assert_that(
+        extraction_order_after_migration.exports.values_list('file_format', flat=True), contains_in_any_order(
```
hixi: Would it make sense to have a newline after the last comma?

osmaxx/excerptexport/migrations/0021_export_file_format_verbose_name.py
das-g: > Why is this migration separately?

Because I noticed that the verbose name was missing only after the data migration `0019_one_export_per_file_format` was already checked in and pushed to GitHub.

> Would it make sense join this change into the `0018_export.py`?

Yeah, I guess. :+1: Actually, we could merge all 3 auto-generated migrations of this branch: `0018_export`, `0020_export_conversion_service_job_id` and `0021_export_file_format_verbose_name`, as the data migration `0019_one_export_per_file_format` should work just as fine on this branch's final models.

osmaxx/job_progress/views.py
das-g: Yes, for #438 

(If it turns out to be not needed anymore for that, I'll remove it while working on #438.)

web_frontend/config/settings/common.py
```diff
@@ -54,6 +54,7 @@
     'osmaxx.version',
 
     'osmaxx.countries',
+    'osmaxx.conversion_api',
```
hixi: Is this or will this be an app, or is it just a python module? Or did treating this as a normal module lead to any unexpected errors?

hixi:
>These migration tests are skipped by default because they make other tests fail (probably due to [Django bug #25251](https://code.djangoproject.com/ticket/25251)). To run (only) them, one may use
```bash
./runtests.py --transactional -k test_0019_one_export_per_file_format
```


I think having this in the Readme or development/testing documentation would be a good idea.

osmaxx/utils/frozendict.py
das-g: > is the noqa for lower_casing a ClassName?
Yes

> And why are we lowercasing this in the first place?
* So the code I copy-pasted below that can work unmodified.
* So we could switch to dictproxyhack by only changing the import line (and `requirements.txt`)
* So it is consistent with `dict()`, `list()` etc.

tests/conftest.py
das-g: Huh, actually I don't need it any more. Needed it when I revived the job_progress tests in b23c04284339a42718f6e70ebc599be8244b1b89, but half of those tests are skipped right now. I'll remove it for now. We'll probably need it again when re-reviving those tests (or replacing them with new ones), but the error message if we forget to re-add the app will be clear enough, IIRC.

das-g: :white_check_mark: done: c49d9b1

tests/excerptexport/migrations/conftest.py
das-g: If I would create the executor in the inner function (`_migrate`), then a new instance would be created for each invocation, even if in the same test-function. I could of course create it in the outer function (`migrate`)  and hold it in a variable there that is accessed by the inner function, but a fixture seemed more elegant to me.

tests/excerptexport/migrations/test_0019_one_export_per_file_format.py
das-g: Dunno. Is
```python
    assert_that(
        extraction_order_after_migration.exports.values_list('file_format', flat=True),
        contains_in_any_order(
            'txt',
            'foobar',
        )
    )
```
more readable? If the line is too long to comfortably comprehend, it might make more sense to extract `extraction_order_after_migration.exports.values_list('file_format', flat=True)` to a local variable and to give that a good name.

web_frontend/config/settings/common.py
hixi: let's just leave this as is for now :-)

tests/excerptexport/migrations/test_0019_one_export_per_file_format.py
hixi: For me it is more readable, because my eyes scan for the line-start. Extracting seems a good alternative as well.

osmaxx/excerptexport/migrations/0021_export_file_format_verbose_name.py
das-g: Like so? f2049a6

Or should I have manually appended the `operations` lists?

das-g:
> I think having this in the Readme or development/testing documentation would be a good idea.

See [my comments](https://github.com/geometalab/osmaxx-frontend/pull/439#commitcomment-17159962) on 03ac356.

PR #444, opened 2016-04-20
fix: pass only SRS instead of option dict to ConversionApiClient.create_parametrization

das-g:
because [that's what `ConversionApiClient.create_parametrization` expects](https://github.com/geometalab/osmaxx-frontend/commit/2e810927dd0544d875f9c7dd92fe6269e682f01a#diff-8c8255674e202573bed3d2495394391eR43).

### Reviewed by
- [ ] @hixi


das-g:
@hixi said in chat:
> looks good to me, no remarks. Can't do anything until tomorrow, since I'm already on my way home :wink:
> go ahead an merge :+1:

PR #445, opened 2016-04-20
change Export.file_format from TextField to CharField

das-g:
### Reviewed by
- [x] @hixi


osmaxx/excerptexport/models/export.py
```diff
@@ -19,7 +19,7 @@ class Export(models.Model):
     """
     extraction_order = models.ForeignKey('ExtractionOrder', related_name='exports',
                                          verbose_name=_('extraction order'))
-    file_format = models.TextField(choices=FORMAT_CHOICES, verbose_name=_('file format / data format'))
+    file_format = models.CharField(choices=FORMAT_CHOICES, verbose_name=_('file format / data format'), max_length=10)
```
hixi: If we ever have a format that has a `short_name` longer than 10 we need a migration just for that, I think it doesn't hurt having a longer length here, like 30 - that looks more or less future safe ;-)

das-g: Wouldn't a new choice require a migration, anyway?

PR #446, opened 2016-04-22
Feature/#438 handle status updates (part 1)

das-g:
First part of #438:

### Implemented in this PR:
- If callback received and status in callback payload different from export model instance status:
  1. update export model instance status
  2. notify user by on-screen message
- Share status definitions between conversion service and frontend

### Still to do in subsequent PRs:
- Query conversion service for jobs of state export statuses
- Decide in more detail what should happen upon what status change (see https://github.com/geometalab/osmaxx-frontend/issues/438#issue-148772137) and implement that accordingly.
- Remove status (field `state`) from extraction order model
- Move downloads from extraction order model to export model

### Reviewed by
- [x] @hixi

### Please note
- Export model statuses are initialized/set independent of extraction order status. So the information about the order's states will be lost when the corresponding field is removed in favor of the new field on the export model.


tests/conftest.py
```diff
@@ -75,6 +75,7 @@ def pytest_configure():
             'rest_framework_gis',
             'rest_framework.authtoken',
             'crispy_forms',
+            'stored_messages',
```
hixi: Are we relying in tests on this app being present? I couldn't find it in the changed code, but that doesn't mean it's not used ;-)

hixi:
very nice! :+1:

tests/conftest.py
das-g: > Are we relying in tests on this app being present? I couldn't find it in the changed code [...]

That would be [the call to](https://github.com/geometalab/osmaxx-frontend/pull/446/files#diff-e52c21d719a322835dc4db89ea595076R58) `osmaxx.utilities.shortcuts.Emissary.info` which uses `stored_messages` [via `Emissary.inform`](https://github.com/geometalab/osmaxx-frontend/blob/0704004e8da31444f947bec022ef60132284224d/osmaxx/utilities/shortcuts.py#L80).

hixi: :+1:

PR #447, opened 2016-04-24
Eliminate unused CANCELED order state

das-g:
There are currently no plans to allow cancellation of jobs. If we ever implement that, this code would quickly be resurrected or re-implemented. Until then, YAGNI.

### Reviewed by
- [x] @hixi


hixi:
:+1:

PR #449, opened 2016-04-25
simple python script to release images on dockerhub

hixi:
Reviewed by:
- [x] @das-g 

Just a simple release script to be able to release images on dockerhub more easily while transitioning.


build_amd_push_images.py
```diff
@@ -0,0 +1,32 @@
...
+
+def docker_build(dockerfile, image_name, release, location='.'):
+    subprocess.check_call(['docker', 'build', '-f', dockerfile, '-t', '{}:{}'.format(image_name, release), location])
+
+
+def docker_push(release, image_name, *args, **kwargs):
+    subprocess.check_call(['docker', 'push', '{}:{}'.format(image_name, release)])
+
+
+def parse_args(args):
```
das-g: This function is never invoked.

build_amd_push_images.py
```diff
@@ -0,0 +1,32 @@
...
+    parser = argparse.ArgumentParser(description='Image pusher for a new OSMaxx release.')
+    parser.add_argument('release', type=str, help='the name of the new release')
+    return parser.parse_args(args)
+
+
+if __name__ == '__main__':
+    import subprocess
+    release = subprocess.check_output(["git", "describe"]).strip().decode()
```
das-g: Consider including the `--dirty` option, so that releases based on uncommitted changes can be identified. (Alternatively, check whether the working dir is clean and refuse operation is it isn't.)

build_and_push_images.py
```diff
@@ -0,0 +1,24 @@
...
+    subprocess.check_call(['docker', 'build', '-f', dockerfile, '-t', '{}:{}'.format(image_name, release), location])
+
+
+def docker_push(release, image_name, *args, **kwargs):
+    subprocess.check_call(['docker', 'push', '{}:{}'.format(image_name, release)])
+
+
+if __name__ == '__main__':
+    import subprocess
+    release = subprocess.check_output(["git", "describe"]).strip().decode()
```
das-g: Consider including the `--dirty` option, so that releases based on uncommitted changes can be identified. (Alternatively, check whether the working dir is clean and refuse operation is it isn't.)

PR #452, opened 2016-04-26
Feature/#438 handle status updates part 2

das-g:
transfer `OutputFile`s from `ExtractionOrder` to `Export`s

### Reviewed by
- [x] @hixi

connected to #438


osmaxx/excerptexport/models/output_file.py
```diff
@@ -5,34 +5,35 @@
...
             'date': self.creation_date.strftime("%F"),
-            'excerpt_name': self.extraction_order.excerpt_name.replace(" ", ""),
+            'excerpt_name': self.export.extraction_order.excerpt_name.replace(" ", ""),
             'content_type': self.content_type if self.content_type else 'file',
             'file_extension': self.file_extension
         }
 
+    @property
+    def file_extension(self):
+        return self.export.file_format
```
hixi: Does this always work? `null=True` seems to be case, where this property would raise an error. I think properties that (could potentially) raise an error are bad design, not?

das-g: Good catch.

After the [data migration](https://github.com/geometalab/osmaxx-frontend/pull/452/files#diff-dd7cd01aee17d8397c83f32ac94d06f3R36), I should disallow `NULL` for `OutputFile.export`.

hixi: Good idea, just add another migration step :+1:

das-g: Done in [`12dbb5d` & `8591a48`](https://github.com/geometalab/osmaxx-frontend/pull/452/files/f3f526766c6e5c97a3c2858368a3db15570912a2..8591a48a30f7c23662c492a0521bd1b8f71853f0)

osmaxx/excerptexport/migrations/0022_transfer_output_files_from_extraction_orders_to_exports.py
```diff
@@ -0,0 +1,53 @@
...
+
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+def set_export_fk_from_extraction_order_and_content_type(apps, schema_editor):
+    OutputFile = apps.get_model('excerptexport', 'OutputFile')  # noqa
+    for output_file in OutputFile.objects.all():
+        output_file.export = output_file.extraction_order.exports.get(file_format=output_file.content_type)
+        output_file.save()
```
das-g: I didn't write automated tests for this data migration. Best would probably be to test it on a **copy** of our production data. Is that feasible?

hixi: Doable, yes, feasible, not so much, otherwise you wouldn't be asking, right?

If you wanted to test this, extract it to a function and test that one, probably easier than getting the data (in the sense of how much time it needs).

osmaxx/excerptexport/models/output_file.py
hixi: :+1:

osmaxx/excerptexport/migrations/0022_transfer_output_files_from_extraction_orders_to_exports.py
das-g: Problem with (only) unit-ish testing this particular migration: That would just proves that it works for the data I assume to be around, but that assumption might be wrong to begin with. I've had a look at https://&lt;production domain>/admin/excerptexport/outputfile/ and that _looks_ as if my assumptions are correct, but I'm unsure whether that's enough.

osmaxx/excerptexport/models/output_file.py
```diff
@@ -5,34 +5,35 @@
...
 
-from .extraction_order import ExtractionOrder
+from osmaxx.excerptexport.models.export import Export
 from osmaxx.excerptexport.utils.upload_to import get_private_upload_storage
 
 
 class OutputFile(models.Model):
     mime_type = models.CharField(max_length=64, verbose_name=_('mime type'))
-    file_extension = models.CharField(max_length=64, verbose_name=_('file extension'), default='')
     content_type = models.CharField(max_length=64, verbose_name=_('content type'), default='')
```
das-g: oops, I should have removed `content_type`, not `file_extension`

das-g:
don't merge for now (see https://github.com/geometalab/osmaxx-frontend/pull/452#discussion_r61260025)

das-g:
replaced by #455

PR #453, opened 2016-04-27
Feature/#435 modify website link buttons

### LINKED ISSUE #435: Modify Website Link buttons
- Remove home button
- move logout to right
- ~~Add special links (impressum/about) to footer~~ can't be done without those pages being available, covered in #434

hixi:
Reviewed by:
- [x] @das-g 

Closes #435


das-g:
Looks good (except for partial ending without newline), but doesn't cover all points of #435:
> Add special links (impressum/about) to footer

is isn't done with this change, so either don't let this PR close #435 or extract that point to a separate issue before merging.

hixi:
Issue updated

Issue #434, opened 2016-04-12
Static About Site


hixi:
- [x] create a static page with information about the OSMaxx project.
- [x] Link to that page in footer.
- [x] Add Impressum section as well.
- [x] Add link and explanation to OsmAnd #380

hixi:
@sfkeller do you know what should/needs to be included in the `about us` section (impressum, information about the team, contact...)?

sfkeller:
Don't forget to add an information where OSMaxx declares which data sources where used besides.

PR #454, opened 2016-04-27
Feature/#434 add static pages

hixi:
Part of #434 

Does some more:
- moved templates and static files to separate app that are not belonging to excerptexports
- splitted base.html into more usual chunks (base.html for styling, osmaxx/base.html for base functionality and excerptexport/base.html for excerptexport specific html)
- added links to footer
- added about us button on top navigation list

Reviewed by:
- [x] @das-g 

Todo after this:
- decide in the content for the static-pages


das-g:
Interesting that [removed `osmaxx/core/static/osmaxx/libraries/bootstrap/css/bootstrap.min.css`](https://github.com/geometalab/osmaxx-frontend/blob/feature/%23434_add_static_pages/osmaxx/core/static/osmaxx/libraries/bootstrap/css/bootstrap.min.css) and [added `osmaxx/excerptexport/static/excerptexport/libraries/bootstrap/css/bootstrap.min.css`](https://github.com/geometalab/osmaxx-frontend/blob/f453a20dc27e5dd6a1a82975201e6f3a349f9e1b/osmaxx/excerptexport/static/excerptexport/libraries/bootstrap/css/bootstrap.min.css) are different, even though both claim to be Bootstrap v3.3.6.

Seems to be only whitespace difference, though, so maybe the new one is just minified more aggressively. Have we minified these ourselves or are these the minified variants offered by upstream?

osmaxx/core/templates/base.html
```diff
@@ -0,0 +1,19 @@
+<!DOCTYPE html>
+{% load staticfiles %}
+
+<html>
+<head lang="en">
+    <meta charset="UTF-8">
+    <title>Osmaxx</title>
+    <link rel="icon" type="image/png" href='{% static "osmaxx/images/favicon.png" %}' />
```
das-g: Would title and favicon belong into `osmaxx/base.html` instead of here in `base.html`? Everything else here looks generic enough to be independent of the OSMaxx project.

If not, what is the rule (of thumb) for what belong here, and what into `osmaxx/base.html`?

osmaxx/core/templates/pages/about_us.html
```diff
@@ -0,0 +1,28 @@
...
+<div class="container-row content">
+    <h2>About OSMaxx</h2>
+    <p>
+        OSMaxx (short for "OpenStreetMap arbitrary excerpt export") cuts out OpenStreetMap data, processes it to
+        geodata, converts it to typical GIS fileformats and offers those for download.
+    </p>
+    <h3>Useful links</h3>
+    <p>Can be found in our separate <a href="{% url 'pages:downloads' %}">downloads page</a></p>
+
+    <h3>Impressum</h3>
```
das-g: "Impressum" is German. English term would be "imprint". Though I'm not sure whether that would only be used in actually _printed_ material.

osmaxx/core/templates/pages/about_us.html
```diff
@@ -0,0 +1,28 @@
...
+        OSMaxx has been developed by the Geometalab at Hochschule Zürich.
+        Geometalab can be reached via email: <a href="mailto:geometalab@hsr.ch">geometalab@hsr.ch</a>.
+    </p><p>
+        <strong>Geometalab</strong><br />
+        HSR Hochschule für Technik Rapperswil<br />
+        Institute for Software<br />
+        c/o Stefan Keller<br />
+        Oberseestrasse 10<br />
+        Postfach 1475<br />
+        CH-8640 Rapperswil<br />
```
das-g: Do we need the postal address here? (I know it's required in Germany, but Swiss publication laws are more liberal AFAIK.) More useful for interested users would probably be links to geometalab, IFS and HSR websites. (On the IFS website they could still find our postal address if desired.)

osmaxx/core/templates/pages/about_us.html
```diff
@@ -0,0 +1,28 @@
...
+    <p>
+        OSMaxx (short for "OpenStreetMap arbitrary excerpt export") cuts out OpenStreetMap data, processes it to
+        geodata, converts it to typical GIS fileformats and offers those for download.
+    </p>
+    <h3>Useful links</h3>
+    <p>Can be found in our separate <a href="{% url 'pages:downloads' %}">downloads page</a></p>
+
+    <h3>Impressum</h3>
+    <p>
+        OSMaxx has been developed by the Geometalab at Hochschule Zürich.
```
das-g: "Hochschule **Zürich**"?

web_frontend/config/urls.py
```diff
@@ -10,4 +10,5 @@
     url(r'^api-auth/', include('rest_framework.urls', namespace='rest_framework')),
     url(r'^api/', include('osmaxx.excerptexport.rest_api.urls', namespace='excerptexport_api')),
     url(r'^job_progress/', include('osmaxx.job_progress.urls', namespace='job_progress')),
+    url(r'^pages/', include('osmaxx.core.urls', namespace='pages')),
```
das-g: I don't like the `pages` namespacing in `osmaxx.example.com/pages/about/` and in `osmaxx.example.com/pages/downloads/`. (In the user-visible URLs. I don't care about the `namespace='pages'`.) Then again, we have to distinguish between the static downloads page and the user's list of downloads and the latter already occupies `osmaxx.example.com/downloads/`.

Or could we combine those two pages? The one-stop<del> shop</del><ins> URL</ins> for OSMaxx downloads, whether static or not. (Obviously with more content when logged in.)

osmaxx/core/templates/base.html
hixi: The general idea is to easily apply a different *base* by providing (read: overriding) the default `base.html` with minimal interference. 
Good idea to move that part down the cascade.

osmaxx/core/templates/pages/about_us.html
hixi: `imprint` is not correct either: http://www.language-boutique.com/lost-in-translation-full-reader/impressum-or-imprint.html

But `legal disclosure` sounds quite reasonable or `company details`, even though `company` might be somewhat misleading.

osmaxx/core/templates/pages/about_us.html
hixi: I know that the swiss laws are more liberal, but having an Adress is actually quite OK - but shorter version would do, I guess ;-).

osmaxx/core/templates/pages/about_us.html
hixi: ooops.

web_frontend/config/urls.py
hixi: We combine them, but if you look at our current download page, we already have a lot of content there when logged in, so I don't think that would that helpful. The `pages` naming isn't great, I agree, but I don't want to have `static_pages` since they are not static for the user, and `cms` is also wrong ;-) .

They are static pages for the user, as the rest, and since we already register the excerpt exports urls[1] globally, that option is out of question as well.

Cleaner approach would be:

- include `excerpt_export` urls in `core.urls`
- inlcude `core.urls` as top level urls
- combine both download pages and push the static downloads to the right side (ie. 4 of 12 columns?) when logged in

[1]: top level registered urls
```
url(r'', include('osmaxx.excerptexport.urls', namespace='excerptexport')),
...
url(r'', include('social.apps.django_app.urls', namespace='social')),
```

hixi:
@das-g interesting indeed, they should be the exact same file, since I only moved them locally...

osmaxx/core/templates/base.html
hixi: done

osmaxx/core/templates/pages/about_us.html
hixi: I went for `company information`

osmaxx/core/templates/pages/about_us.html
hixi: now shorter, better?

das-g: Oops, not as liberal as I thought they'd be. [UWG Art. 3 "Unlautere Werbe- und Verkaufsmethoden und anderes widerrechtliches Verhalten"](https://www.admin.ch/opc/de/classified-compilation/19860391/index.html#a3):
> Unlauter handelt insbesondere, wer:
> [...]
> **s.** Waren, Werke oder Leistungen im elektronischen Geschäftsverkehr anbietet und es dabei unterlässt:
>
> 1. klare und vollständige Angaben über seine Identität und seine Kontaktadresse einschliesslich derjenigen der elektronischen Post zu machen,
> 2. auf die einzelnen technischen Schritte, die zu einem Vertragsabschluss führen, hinzuweisen,
> 3. angemessene technische Mittel zur Verfügung zu stellen, mit denen Eingabefehler vor Abgabe der Bestellung erkannt und korrigiert werden können,
> 4. die Bestellung des Kunden unverzüglich auf elektronischem Wege zu bestätigen;

We offer either "Werke" or "Leistungen" and even though the contract is not made online (so point 2 doesn't apply) we probably do so "im elektronischen Geschäftsverkehr".

web_frontend/config/urls.py
das-g: Let's keep it as-is-in-this-PR for now and discuss this again when the (dynamic) downloads page gets revised anyway (due to focusing on excerpts rather than orders).

das-g: The column idea for combining them sounds good, though.

osmaxx/core/templates/pages/about_us.html
das-g: > now shorter, better?

Actually, if we do keep the address (and IANAL, TINLA, but I guess the law says we must), let's use the postal address as it should be used on envelopes when sending snail mail to us. I guess that corresponds to the longer address as it was before 5b1e738d35be4671e89566f843737a70dac6c31e.

That is, unless HSR has some guidelines on what to use for impressums. @sfkeller do you know whether such guidelines exist?

> I went for `company information`

What if we call the page just "about" and this section "About Us"? That would go well with the "About OSMaxx" section above it.

hixi:
@das-g after the changes, is it OK if I merge this?

osmaxx/core/templates/pages/about_us.html
```diff
@@ -24,5 +25,23 @@
         Postfach 1475<br />
         CH-8640 Rapperswil<br />
     </p>
+
+    <h3>Data Information</h3>
+    <h4>Base Data</h4>
+    <p>
+        All the base data used by OSMaxx stems originally from
+        <em>OSM</em> and can be used under their terms.
+    </p>
```
das-g: I think I'd go for something like

> ### Data sources
> All data used by OSMaxx comes directly or indirectly from [OpenStreetMap](https://www.openstreetmap.org/):
> #### Main data source
> The bulk of our input data (everything not listed separately below) is verbatim OpenStreetMap data, taken from our own regularly synchronized database import of the [OpenStreetMap planet](http://planet.openstreetmap.org/).
> 
> #### Additional data sources
> ##### Coastlines, [...]
> [...]
> ##### Country Boundaries
> [...]

osmaxx/core/templates/pages/about_us.html
```diff
@@ -24,5 +25,23 @@
...
+    <h4>Coastlines, Waterways and Landmass</h4>
+    <p>
+        Coastlines, waterways and landmass are from
```
das-g: water**ways** (that is rivers etc.), too? I thought this was just for the sea.

osmaxx/core/templates/pages/about_us.html
hixi: Feel free to change right in here :-)

osmaxx/core/templates/pages/about_us.html
hixi: only the sea, my mistake.

hixi: corrected.

osmaxx/core/templates/pages/about_us.html
das-g: Done partially in 838c24d2a330ed5517d688925d23355e8b6463f7
let's add the "License" section later

hixi: :+1:

hixi:
@das-g mergeable state? then LGTM ;-)

PR #455, opened 2016-04-27
Feature/#438 handle status updates part 2

das-g:
transfer `OutputFile`s from `ExtractionOrder` to `Export`s
(Replaces PR #452)

### Reviewed by
- [x] @hixi

connected to #438


osmaxx/excerptexport/migrations/0022_transfer_output_files_from_extraction_orders_to_exports.py
```diff
@@ -0,0 +1,53 @@
...
+        migrations.AddField(
+            model_name='outputfile',
+            name='export',
+            field=models.OneToOneField(null=True, on_delete=django.db.models.deletion.CASCADE, related_name='output_file', to='excerptexport.Export', verbose_name='export'),
+        ),
+        migrations.RunPython(
+            set_export_fk_from_extraction_order_and_content_type,
+            set_extraction_order_fk_and_content_type_from_export
+        ),
+        migrations.AlterField(
```
hixi: :+1:

tests/excerptexport/test_downloads.py
```diff
@@ -24,7 +24,8 @@ def test_file_download(self, *args):
         excerpt = Excerpt.objects.create(name='Neverland', is_active=True, is_public=True, owner=user,
                                          bounding_geometry=bg)
         extraction_order = ExtractionOrder.objects.create(excerpt=excerpt, orderer=user)
-        output_file = OutputFile.objects.create(mime_type='test/plain', extraction_order=extraction_order,
+        export = extraction_order.exports.create(file_format='fgdb')
+        output_file = OutputFile.objects.create(mime_type='test/plain', export=export,
```
hixi: am I right to assume `test/plain` is no typo?

das-g: I didn't change that in this PR.

I guess @wasabideveloper and I placed some dummy value for test purposes back when this test was written. Quite possible we meant it to be <code>te**x**t/plain</code> (MIME type for plain text), though. I don't think it matters.

hixi: It doesn't, I was just curious :-)

das-g: I wonder though whether we should persist the MIME type at all, seeing that we only put hard-coded values into that field. We could just as well hard-code the value in the view that delivers the file and currently uses the field. Or, if we want to be more flexible, determine the MIME type dynamically (either when persisting the file (and store it in the existing field), or when delivering it) either from the filename extension using the [`mimetypes`](https://docs.python.org/3/library/mimetypes.html) module or from the file content using `/usr/bin/file --mime` (which uses libmagic) or any of the several [python wrappers of libmagic](http://stackoverflow.com/a/12297929/674064).

wasabideveloper: The result files are zipped, right? I wasn't able to determine the type of format by the file because our zips are of a custom structure and containing directories and somwhere in there some geo format. So the easiest way to do this was to hard code the mime type in the process receiving the geo file from the converter. Because at this point, we did know which format is inside the zip. But I'm not shure if this problem already exist about the new architecture.

wasabideveloper: I don't remember what we thought about `test/plain`. I think it's a typo copied two times.

hixi:
@das-g Can I merge this?

tests/excerptexport/test_downloads.py
hixi: We could also use this: http://django-downloadview.readthedocs.io/en/1.9/

PR #457, opened 2016-04-28
Replace tabs with spaces in sql files 4 spaces

hixi:
Use 4 Spaces and keep comments aligned.
Reviewed by:
- [ ] @das-g


hixi:
wrong branch.

PR #459, opened 2016-04-28
Feature/ready combine for production

hixi:
Previewed by:
- [x] @das-g 

Only for preview. Still some work needs to be done.

These changes where necessary to get it running with frontend/backend integration.

Done:
- download files again
- get some files when they are available

Don't merge this, failing tests are on purpose!


das-g:
What is my task with this PR?

> **P**reviewed by:
> - [ ] @das-g 
> 
> Only for **p**review. Still some work needs to be done.

(Emphasis mine)
Is the goal of the **p**review a review?

osmaxx/excerptexport/models/export.py
```diff
@@ -27,15 +29,16 @@ class Export(models.Model):
...
 
     def send_to_conversion_service(self, clipping_area_json, incoming_request):
         from osmaxx.api_client.conversion_api_client import ConversionApiClient
         api_client = ConversionApiClient()
         extraction_format = self.file_format
         gis_options = self.extraction_order.extraction_configuration['gis_options']
         out_srs = gis_options['coordinate_reference_system']
-        parametrization_json = api_client.create_parametrization(clipping_area_json, extraction_format, out_srs)
-        job_json = api_client.create_job(parametrization_json, self.status_update_url)
+        parametrization_json = api_client.create_parametrization(boundary=clipping_area_json, out_format=extraction_format, out_srs=out_srs)
```
das-g: :+1:

osmaxx/api_client/conversion_api_client.py
```diff
@@ -129,6 +129,18 @@ def _create_job_TODO_replace_me(self, extraction_order, request):  # noqa
...
+    def get_result_file_url(self, job_id):
+        job_detail_url = CONVERSION_JOB_URL + '{}/'.format(job_id)
+        return self.authorized_get(job_detail_url).json()['resulting_file']
+
+    def download_file_to(self, url, destination_filename):
+        r = self.authorized_get(url, stream=True)
+        with open(destination_filename, 'wb') as f:
+            for chunk in r.iter_content(chunk_size=1024):
+                if chunk:
+                    f.write(chunk)
```
das-g: And is there any benefit to this manual chunking?

And why is the file written to disk here and read from disk and written again [in the caller](https://github.com/geometalab/osmaxx-frontend/pull/459/files#diff-e52c21d719a322835dc4db89ea595076R67)? Wouldn't it make more sense to pass it in-memory (e.g. as `ContentFile`)?

hixi: ## Chunking
Our files can be quite big, serving them in one load (loading them into memory) isn't a good idea. A much better solution would be to use NGINX's `X-SENDFILE`; something like [django-sendfile](https://pypi.python.org/pypi/django-sendfile) or [django-downloadview](https://django-downloadview.readthedocs.io/en/1.9/), see also an [older blogpost](http://www.sensibledevelopment.com/2010/11/django-sendfile-an-for-abstraction-large-file-serving-in-django/)

## Saving the file "again"

That's how this [must](http://stackoverflow.com/questions/7514964/django-how-to-create-a-file-and-save-it-to-a-models-filefield) be, unfortunately.

See also [FieldFile](https://docs.djangoproject.com/en/1.9/ref/models/fields/#django.db.models.fields.files.FieldFile.save) and [FileField](https://docs.djangoproject.com/en/1.9/ref/models/fields/#filefield-and-fieldfile)

hixi:
All Code in here has been merged in a one form or another, AFAIK. Closing.

das-g:
> All Code in here has been merged in a one form or another, AFAIK. Closing.

[The change to `osmaxx/excerptexport/templates/excerptexport/templates/extraction_order_status.html`](https://github.com/geometalab/osmaxx-frontend/pull/459/files#diff-a1de77418ae20144f658fdac81759d55R34) hasn't, as far as I can see.

PR #461, opened 2016-05-02
Feature/#438 handle status updates

### LINKED ISSUE #438: let front-end properly react to conversion job status updates from revamped conversion service

das-g:
closes #438

re-implemented pre-API-refactoring functionality like:

- Send emails!
- Actually fetch results from the conversion service!
- Fancy colors!

### Reviewed by
- [x] @hixi


osmaxx/api_client/conversion_api_client.py
```diff
@@ -70,6 +71,15 @@ def create_job(self, parametrization, callback_url):
         response = self.authorized_post(url='conversion_job/', json_data=json_payload)
         return response.json()
 
+    def get_result_file(self, job_id):
+        download_url = self._get_result_file_url(job_id)
+        response = self.authorized_get(download_url)
+        return ContentFile(response.content)
```
hixi: ContentFile keeps the file in memory. They can be quite big, especially since we will be allowing bigger areas to be cut out. I'd favor a solution that doesn't write the file into memory.

osmaxx/api_client/conversion_api_client.py
```diff
@@ -114,7 +124,7 @@ def _create_job_TODO_replace_me(self, extraction_order, request):  # noqa
             }
         })
         try:
-            response = self.authorized_post(CONVERSION_JOB_URL, json_data=request_data)
+            response = self.authorized_post(OLD_CONVERSION_JOB_URL, json_data=request_data)
```
hixi: I guess this is a temporary solution to not make tests fail? Because this doesn't `actually` create a job on the conversion-service ;-)

osmaxx/excerptexport/models/export.py
```diff
@@ -47,16 +51,23 @@ def get_full_status_update_uri(self, request):
     def status_update_url(self):
         return reverse('job_progress:tracker', kwargs=dict(export_id=self.id))
 
-    def set_and_handle_new_status(self, new_status):
+    def set_and_handle_new_status(self, new_status, *, incoming_request):
         if self.status != new_status:
```
hixi: I know this is not part of this pull-request: what happens, if the passed in status is `None`? (It will fail, since null=False). This will make all subsequent calls django fail because of our middleware calling this function. Should that be filed as a separate bug?

osmaxx/excerptexport/models/export.py
```diff
@@ -65,3 +76,23 @@ def _get_export_status_changed_message(self):
...
             context=view_context,
         ).strip()
+
+    def _fetch_result_file(self):
+        from osmaxx.api_client import ConversionApiClient
+        from . import OutputFile
+        api_client = ConversionApiClient()
+        file_content = api_client.get_result_file(self.conversion_service_job_id)
+        file_location = get_default_private_storage().save(
+            name=str(uuid.uuid4()),
```
hixi: This should probably be a better name than this, something with an UUID inside it's name would be great, but also some more information like the file_format, its name (from the order) maybe more?

osmaxx/excerptexport/models/extraction_order.py
```diff
@@ -146,6 +147,38 @@ def get_absolute_url(self):
         from django.core.urlresolvers import reverse
         return reverse('excerptexport:status', kwargs={'extraction_order_id': self.id})
 
+    def send_email_if_all_exports_done(self, incoming_request):
+        if all(export.is_status_final for export in self.exports.all()):
```
hixi: :+1: I like the expression :-)

osmaxx/excerptexport/models/export.py
das-g: > Should that be filed as a separate bug?
No, let's fix it in this PR.

osmaxx/excerptexport/templates/excerptexport/email/all_exports_of_extraction_order_done_subject.txt
```diff
@@ -0,0 +1 @@
+Extraction Order #{{ extraction_order.id }} "{{ extraction_order.excerpt_name }}": {{ successful_exports_count }} Export{{ successful_exports_count|pluralize }} finished successfully, 1 failed
```
hixi: I don't think there will always be `1 failed`, so maybe remove that hard coded part :-)

osmaxx/excerptexport/models/export.py
das-g: Fixed in a sub-PR (i.e., one that'd merge into `feature/#438-handle_status_updates`): #462

osmaxx/api_client/conversion_api_client.py
das-g: Yes, `_create_job_TODO_replace_me` should eventually be removed, once we've switched to `extraction_order.forward_to_conversion_service` in the view.

osmaxx/excerptexport/models/export.py
das-g: Does the conversion service already produce a useful filename? If so, we could re-use that.

osmaxx/excerptexport/models/extraction_order.py
das-g: I dislike how grammatically wrong it is when read as English :-P

But yeah, otherwise, it's quite neat. :smile:

hixi: correct english is quite high expectations for a programming language :laughing:

osmaxx/excerptexport/models/export.py
hixi: Yes, [somewhat](https://github.com/geometalab/osmaxx-frontend/blob/combine_repos/osmaxx/conversion/models.py#L80), but I think that would probably be a great idea to keep that one or at least change it there.

das-g: Looking at the code, it seems to be quite different for Garmin than for the PostGIS/ogr2ogr-based exports.

Should we improve that service-side or build our own, consistent zip file naming scheme in the frontend?

hixi: I think improving it on the backend would be better, having the naming at one place only seems more sensible to me.

osmaxx/excerptexport/templates/excerptexport/email/all_exports_of_extraction_order_done_subject.txt
das-g: oops, missed to parametrize completely that after copy-pasting the expected string for the test

now done in 3819fdecc2450c3419602b4323fa7a6f6f576fdc

osmaxx/api_client/conversion_api_client.py
das-g: Now more chunked (if > 10 KB): https://github.com/geometalab/osmaxx-frontend/pull/461/files#diff-8c8255674e202573bed3d2495394391eR74

osmaxx/api_client/API_client.py
```diff
@@ -90,5 +91,17 @@ def _authorization_headers(self):
         }
 
 
+class LazyChunkedRemoteFile:
+    def __init__(self, *args, download_fu=requests.get, **kwargs):
```
hixi: Maybe document why this works, otherwise we ask ourselves this question next time ;-)

osmaxx/api_client/conversion_api_client.py
```diff
@@ -73,8 +73,7 @@ def create_job(self, parametrization, callback_url):
 
     def get_result_file(self, job_id):
         download_url = self._get_result_file_url(job_id)
-        response = self.authorized_get(download_url)
-        return ContentFile(response.content)
+        return LazyChunkedRemoteFile(download_url, download_fu=self.authorized_get)
```
hixi: fu = `foo` -> `download_function` might be better in this particular case.

osmaxx/api_client/API_client.py
das-g: done: 289de78

osmaxx/api_client/conversion_api_client.py
das-g: done: abae62c

osmaxx/api_client/API_client.py
```diff
@@ -90,5 +91,28 @@ def _authorization_headers(self):
...
 
+class LazyChunkedRemoteFile:
+    """Wrapper around to-be-downloaded files
+
+    Can be passed to django.core.files.storage.Storage#save as ``content``
+    if that storage's save logic consumes chunks like ``FileSystemStorage`` does
+    at https://github.com/django/django/blob/1.9.4/django/core/files/storage.py#L252
+
+    Notes:
+        This is not a file-like object: It has no ``read`` method.
```
das-g: A `read` method could probably provided by delegating to [`response.raw.read()`](http://docs.python-requests.org/en/v2.8.1/api/#requests.Response.raw), but I haven't tested that and for now YAGNI.

osmaxx/api_client/API_client.py
```diff
@@ -92,9 +92,23 @@ def _authorization_headers(self):
...
+    Notes:
+        - This is not a file-like object: It has no ``read`` method.
+        - Object initialization opens the connection for downloading, but will not fetch the content until the chunks
+          are requested. If too much time has passed between those steps, the server might have closed the connection,
+          so don't wait too long.
+    """
+
+    def __init__(self, *args, download_function=requests.get, **kwargs):
+        stream = kwargs.pop('stream', True)
+        assert stream, "can't chunk without streaming"
```
hixi: Just checking if this is the desired behavior: This will only raise if stream is explicitly set to False, if it isn't set, it will set it to true and work.

das-g: Yes, that's the intended behavior. To do something useful in this class, we require streaming, so
- there's a conflict if the caller explicitly requests `stream=False`. We can't fulfill that wish and still do something useful, so we raise
- if the caller doesn't pass the `stream` parameter, we assume they don't care and set it to `True`, as that's what we need
- if the caller passes `stream=True`, there's no conflict and we use that
- if the caller passes `stream=<something truthy, but not True>`, we use that, too and hope that `download_function` can handle it

PR #462, opened 2016-05-03
Avoid integrity error in middleware

das-g:
Sub-PR of #461 for fixing a bug [found in review](https://github.com/geometalab/osmaxx-frontend/pull/461/files/072bfdb994d2eba90887fa010521cabc3a81949e#r61759916)
(Therefore connected to #438)

### Reviewed by
- [x] @hixi


osmaxx/excerptexport/models/export.py
```diff
@@ -49,6 +49,7 @@ def status_update_url(self):
         return reverse('job_progress:tracker', kwargs=dict(export_id=self.id))
 
     def set_and_handle_new_status(self, new_status, *, incoming_request):
+        assert new_status in dict(STATUS_CHOICES)
```
hixi: `STATUS_CHOICES` could be a nested tuple, in theory, but it practices, this is OK, since we know it isn't, and if it ever should be, we will see this assertion fail :+1:

Issue #329, opened 2015-12-21
freehand polygons as excerpt area


hixi:
This is just a rough overview of what needs to be done to achieve _arbitrary_ (multi) polygons:

Web Frontend:
- [ ] integrate [leaflet.draw](https://github.com/Leaflet/Leaflet.draw)
    - [ ] use custom icon
    - [ ] allow freehand polygons to be drawn

Excluding:
Conversion Service (geometalab/osmaxx-conversion-service#96):
- [x] take an arbitrary shape (Multipolygon) as cutting-area
- [x] generate a `.poly`-file 'on the fly'

PR #472, opened 2016-05-06
Feature/#329 freehand polygons as excerpt area

hixi:
Reviewed by:
- [x] @das-g 

Todo: 
- Rerecord test fixtures, but they are quite big, so I'd rather do this separately


das-g:
1fd4dca & 3fd1fb8 are more related to #467 than #329, aren't they?

hixi:
Yes and no: they are related, but do not exactly implement what is expected there. This is just to get information on the general size and limiting the extraction area (again) to about 5GB.

osmaxx/excerptexport/static/excerptexport/scripts/maps/draw_controls.js
```diff
@@ -0,0 +1,129 @@
...
+            nameField.parentNode.appendChild(e);
+        }
+
+        estimateSize(layer).done(function (data) {
+            function formatBytes(bytes) {
+               if(bytes == 0) return '0 Byte';
+               var k = 1000;
+               var sizes = ['Bytes', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB'];
+               var i = Math.floor(Math.log(bytes) / Math.log(k));
+               return parseFloat((bytes / Math.pow(k, i)).toFixed()) + ' ' + sizes[i];
```
das-g: Avoid a potential (but probably unlikely for quite some time) index out-of-bounds error by using
```js
var i = Math.max(Math.floor(Math.log(bytes) / Math.log(k)), sizes.length - 1);
```

osmaxx/excerptexport/static/excerptexport/scripts/maps/draw_controls.js
```diff
@@ -0,0 +1,129 @@
...
+            e.id = 'error_size_estimation_too_large';
+            nameField.parentNode.appendChild(e);
+        }
+
+        estimateSize(layer).done(function (data) {
+            function formatBytes(bytes) {
+               if(bytes == 0) return '0 Byte';
+               var k = 1000;
+               var sizes = ['Bytes', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB'];
+               var i = Math.max(Math.floor(Math.log(bytes) / Math.log(k)));
```
das-g: Uh, `max` alone won't future-prove this. [The `, sizes.length - 1` part](https://github.com/geometalab/osmaxx-frontend/pull/472#discussion-diff-62335502) would be quite essential ;-)

docker_entrypoint/osmaxx/conversion_service/entrypoint.sh
```diff
@@ -4,7 +4,7 @@ set -e
 # wait for at most 30s for the db to be up
 ${HOME}/entrypoint/wait-for-it.sh ${DATABASE_HOST}:${DATABASE_PORT} -t 30
 
-python3 osmaxx_conversion_service/manage.py migrate && \
+python3 osmaxx_conversion_service/manage.py migrate --no-input && \
```
das-g: Will this make migrations that would otherwise ask for input fail or assume some default answers?

osmaxx/excerptexport/forms/excerpt_form.py
```diff
@@ -30,30 +37,43 @@ def __init__(self, *args, **kwargs):
...
+            # need to add the hidden field validation error to a visible dom
+            self.add_error(None, _('Area Invalid or none selected.'))
+        return clean
+
+    def clean_bounding_geometry(self):
+        allowed_types = ['Polygon', 'Multipolygon']
+        data = self.cleaned_data['bounding_geometry']
+        data_dict = json.loads(data)
+        if data_dict.get('type') not in allowed_types:
+            raise ValidationError(_('Only {} are allowed as input.'.format(allowed_types)))
```
das-g: Translation (`_(...)`) should probably be done before formatting (`... .format(...)`), so that the translations can stay parametrized, too.

Also, this will produce the string
>  Only ['Polygon', 'Multipolygon'] are allowed as input.

i.e., with the python string representation of the array. Is that really what you want?

osmaxx/excerptexport/forms/excerpt_form.py
```diff
@@ -30,30 +37,43 @@ def __init__(self, *args, **kwargs):
...
+        if data_dict.get('type') == 'Polygon':
+            data_dict['coordinates'] = [data_dict.get('coordinates')]
```
das-g: Is there an advantage to use `get` here instead of indexing syntax?

osmaxx/excerptexport/static/excerptexport/scripts/maps/draw_controls.js
hixi: oops, I meant to copy the entire expression.

osmaxx/excerptexport/static/excerptexport/scripts/maps/draw_controls.js
hixi: oops, I meant to copy the entire expression.

docker_entrypoint/osmaxx/conversion_service/entrypoint.sh
hixi: this will just not ask any questions and always to the sane thing: not delete any stale content-types.

osmaxx/excerptexport/forms/excerpt_form.py
hixi: good catch, I copied this a bit too blindly.

osmaxx/excerptexport/forms/excerpt_form.py
hixi: only readability: I thought that 
`data_dict['coordinates'] = [data_dict.get('coordinates')]`
is more easy to read than
`data_dict['coordinates'] = [data_dict['coordinates']]`

(too many parenthesis).

osmaxx/excerptexport/migrations/0025_move_bounding_geometry_20160503_1639.py
```diff
@@ -0,0 +1,33 @@
...
+
+from django.contrib.gis.geos import Polygon, MultiPolygon
+from django.db import migrations
+
+
+def multi_polygon_from(bounding_geometry_old):
+    bb = bounding_geometry_old
+    west, south, east, north = bb.south_west[0], bb.south_west[1], bb.north_east[0], bb.north_east[1]
+    polygon = Polygon.from_bbox((west, south, east, north))
+    return MultiPolygon(polygon)
```
das-g: :+1:

osmaxx/excerptexport/migrations/0025_move_bounding_geometry_20160503_1639.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def move_bounding_geometry_data_to_new_multipolygon_field(apps, schema_editor):
+    Excerpt = apps.get_model('excerptexport', 'Excerpt')  # noqa
+    BBoxBoundingGeometry = apps.get_model('excerptexport', 'BBoxBoundingGeometry')  # noqa
+    for excerpt in Excerpt.objects.all():
+        excerpt.bounding_geometry = multi_polygon_from(
+            BBoxBoundingGeometry.objects.get(pk=excerpt.bounding_geometry_old.id)
```
das-g: Isn't `BBoxBoundingGeometry.objects.get(pk=excerpt.bounding_geometry_old.id)` the same as `excerpt.bounding_geometry_old`?

osmaxx/excerptexport/migrations/0028_move_country_to_excerpt_20160506_0647.py
```diff
@@ -0,0 +1,48 @@
...
+from django.db import migrations
+
+
+def move_country_from_extraction_order_to_export(apps, schema_editor):
+    Excerpt = apps.get_model('excerptexport', 'Excerpt')  # noqa
+    ExtractionOrder = apps.get_model('excerptexport', 'ExtractionOrder')  # noqa
+    Country = apps.get_model('countries', 'Country')  # noqa
+
+    for extraction_order in ExtractionOrder.objects.all():
+        if extraction_order.country_id is not None:
```
das-g: Would
```python
for extraction_order in ExtractionOrder.objects.filter(country_id__is_null=False):
```
do the same?

osmaxx/excerptexport/migrations/0025_move_bounding_geometry_20160503_1639.py
hixi: Yes, it should, but I wasn't sure if the in Memory-Model does that correctly, and instead of a test I just used the safest way, which will fail as loud as possible if something is off..

osmaxx/excerptexport/models/excerpt.py
```diff
@@ -1,24 +1,34 @@
...
 
 class Excerpt(models.Model):
     name = models.CharField(max_length=128, verbose_name=_('name'))
     is_public = models.BooleanField(default=False, verbose_name=_('is public'))
     is_active = models.BooleanField(default=True, verbose_name=_('is active'))
-
     owner = models.ForeignKey(User, related_name='excerpts', verbose_name=_('owner'))
-    bounding_geometry = models.OneToOneField('BoundingGeometry', verbose_name=_('bounding geometry'))
+    bounding_geometry = models.MultiPolygonField(verbose_name=_('bounding geometry'), null=True)
+    country = models.ForeignKey('countries.Country', verbose_name=_('Country'), null=True)
```
das-g: :+1: for using a ForeignKey, now that countries are in the same DB.

Do we want the default `on_delete=CASCADE` here, too, or would `RESTRICT` be more adequate?

osmaxx/excerptexport/templates/excerptexport/base.html
```diff
@@ -12,6 +11,15 @@
...
 
 {% block scripts %}
-<!-- Helpers -->
-<script type="text/javascript" src="{% static 'excerptexport/scripts/helperFunctions.js' %}"></script>
-{% endblock %}
\ No newline at end of file
+    <!-- Helpers -->
+    <script type="text/javascript" src="{% static 'excerptexport/scripts/helperFunctions.js' %}"></script>
+
+    {# MAP SCRIPTS #}
```
das-g: Why use a template comment for naming this section, but HTML comments for naming the other ones?

osmaxx/excerptexport/views.py
```diff
@@ -31,7 +30,7 @@
 
 
 def execute_converters(extraction_order, request):
-    ConversionApiClient()._create_job_TODO_replace_me(extraction_order, request=request)
+    extraction_order.forward_to_conversion_service(incoming_request=request)
```
das-g: :tada: :balloon:

web_frontend/config/urls.py
```diff
@@ -12,3 +14,7 @@
     url(r'^job_progress/', include('osmaxx.job_progress.urls', namespace='job_progress')),
     url(r'^pages/', include('osmaxx.core.urls', namespace='pages')),
 ]
+
+if settings.DEBUG:
+    urlpatterns += static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT, show_indexes=True) +\
+        static(settings.STATIC_URL, document_root=settings.STATIC_ROOT, show_indexes=True)
```
das-g: `static()` returns an empty collection unless in debug mode, so you can unconditionally append it.

osmaxx/excerptexport/migrations/0028_move_country_to_excerpt_20160506_0647.py
hixi: Yes, of course, very good catch.

osmaxx/excerptexport/models/excerpt.py
hixi: I think `on_delete=CASCADE` is actually the correct behavior: We would have to take the boundary from the country and add it as `bounding_geometry` which can't be achieve without using triggers or similar mechanisms. Otherwise we would have invalid objects in the database, which have neither a country nor a `bounding_geometry` set.

osmaxx/excerptexport/templates/excerptexport/base.html
hixi: Because I copied the old code. They should probably all be template comments, they are only developer comments.

docker_entrypoint/osmaxx/conversion_service/entrypoint.sh
hixi: there is no need to do it my way :-)

das-g: > there is no need to do it my way :-)

What hasn't to be done what way? Is this about `manage.py migrate --no-input` or was that a comment for a different line?

hixi: yes, for another comment, I shouldn't answer comments in a comatose state ;-)

osmaxx/excerptexport/migrations/0025_move_bounding_geometry_20160503_1639.py
hixi: there is no need to do it my way :-)

osmaxx/excerptexport/models/excerpt.py
das-g: `on_delete=RESTRICT` would not hamper consistency. It would mean that the DB would refuse to delete a country as long as it is referenced by at least one excerpt.

`CASCADE` means that deleting a country will implicitly delete all excerpts referencing it.

PR #473, opened 2016-05-07
Avoid HTML-escaped plain text

das-g:
### Reviewed by
- [x] @hixi


osmaxx/job_progress/templates/job_progress/messages/export_status_changed.unsave_text
```diff
@@ -1 +1,3 @@
+{% autoescape off %}
 Export #{{ export.id }} "{{ export.extraction_order.excerpt_name }}" to {{ export.get_file_format_display }} has {% if not export.is_status_final %}been {% endif %}{{ export.status }}.
+{% endautoescape %}
```
das-g: We should probably wrap all our plain-text generating (or more generally, not-html-or-xml-generating) templates like this.

hixi: agreed.

PR #488, opened 2016-05-12
update coverage configuration exclusions

hixi:
Reviewed by:
- [x] @das-g 

We shouldn't exclude anything, besides the tests themselves.


das-g:
What happened here? The "Details" links to the CI builds seem to be dead:
> There was an error while loading data.

hixi:
> There was an error while loading data.
No idea.

Issue #466, opened 2016-05-04
Larger downloads fail when downloaded


hixi:
- set larger timeouts
- maybe chunked download, so that larger downloads are possible

hixi:
Also:

- Use nginx as in-between-proxy with config to download the excerpts (and therefore getting rid of this issue altogether)

This would replace the frontend acting as frontend and add a new (lightweight) nginx container acting as proxy, that can also be used easily during testing (on a separate port)

PR #503, opened 2016-05-23
Feature/#466 bugfix allow larger downloads

hixi:
Add new NGINX-Container as Frontend:
- Use standard nginx with obscured path to file as security measure
- disable directory listing (is default, but setting it explicitly here)

Enable using new path:
- move existing files
- redirect download path there

Reviewed by:
- [x] @das-g 

Re-Reviewed by:
- [x] @das-g 

Part of and thus connected to #466


osmaxx/excerptexport/migrations/0030_auto_20160517_1339.py
```diff
@@ -0,0 +1,20 @@
...
+
+    dependencies = [
+        ('excerptexport', '0029_remove_extractionorder_country_id'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='export',
+            name='file_format',
+            field=models.CharField(choices=[('fgdb', 'ESRI File Geodatabase (without coastlines, land and sea polygons)'), ('shapefile', 'ESRI Shapefile'), ('gpkg', 'GeoPackage'), ('spatialite', 'SpatiaLite'), ('garmin', 'Garmin navigation & map data')], max_length=10, verbose_name='file format / data format'),
```
das-g: Doesn't this re-introduce the old temporary fgdb display string from before 32f9dfd49ebfdc5aa54725322c9fb01aa3ee02a3?

osmaxx/excerptexport/models/output_file.py
```diff
@@ -9,11 +9,18 @@
 from osmaxx.excerptexport.utils.upload_to import get_private_upload_storage
 
 
+def uuid_directory_path(instance, filename):
+    # file will be uploaded to MEDIA_ROOT/osmaxx/<public_uuid>/<filename>
+    return '{0}/{1}/{2}/{3}'.format('osmaxx', 'outputfiles', instance.public_identifier, filename)
```
das-g: wouldn't `os.path.join('osmaxx', 'outputfiles', instance.public_identifier, filename)` be more appropriate?

osmaxx/excerptexport/models/output_file.py
```diff
@@ -9,11 +9,18 @@
...
+
+
 class OutputFile(models.Model):
     mime_type = models.CharField(max_length=64, verbose_name=_('mime type'))
     file_extension = models.CharField(max_length=64, verbose_name=_('file extension'), default='')
-    file = models.FileField(storage=get_private_upload_storage(), blank=True, null=True,
-                            verbose_name=_('file'))
+    file_old = models.FileField(storage=get_private_upload_storage(), blank=True, null=True,
+                                verbose_name=_('file'))
```
das-g: wasn't this meant to be removed after the intermediary migrations?

osmaxx/excerptexport/models/output_file.py
```diff
@@ -40,6 +47,11 @@ def __str__(self):
             + ('file: ' + os.path.basename(self.file.name) + ', ' if (self.file and self.file.name) else '') \
             + 'identifier: ' + str(self.public_identifier)
 
+    def get_filename(self):
+        if self.file:
+            return self.file.name.split('/')[-1]
```
das-g: `os.path.basename(self.file.name)` would do the same, I think.

osmaxx/excerptexport/models/output_file.py
```diff
@@ -40,6 +47,11 @@ def __str__(self):
             + ('file: ' + os.path.basename(self.file.name) + ', ' if (self.file and self.file.name) else '') \
             + 'identifier: ' + str(self.public_identifier)
 
+    def get_filename(self):
```
das-g: As this returns an empty string when `self.file` is `None`, rename it to `get_filename_display` or similar. (I'd expect a method named `get_filename` to return `None` or raise an Error if there is no file.)

osmaxx/excerptexport/migrations/0030_auto_20160517_1339.py
hixi: very good catch, I forgot about this migration, should be removed and then the merge can be removed as well, yay :+1:

osmaxx/excerptexport/models/output_file.py
hixi: yes, it should have been removed, thanks!

tests/excerptexport/conftest.py
```diff
@@ -26,27 +26,27 @@ def extraction_order(excerpt, user, db):
...
 def output_file(export):
-    return OutputFile.objects.create(
-        mime_type='test/plain', export=export, content_type='text', file_extension='zip'
-    )
+    return OutputFile.objects.create(mime_type='test/plain', export=export, content_type='text', file_extension='zip')
+
+
+@pytest.fixture
+def output_file_filename():
+    return 'download_file.zip'
```
das-g: :+1:

osmaxx/excerptexport/migrations/0001_squashed_0016_extractionorder_process_start_time.py
```diff
@@ -0,0 +1,143 @@
...
+import django.contrib.gis.db.models.fields
+from django.core.files.storage import FileSystemStorage
+from django.db import migrations, models
+import django.db.models.deletion
+import uuid
+
+
+class PrivateFileSystemStorage(FileSystemStorage):
+    # This is a small hack to still be able to use the standard FileSystemStorage
+    # without having the hardcoded PRIVATE_MEDIA_ROOT in automatic generated migrations.
```
das-g: Is this comment still applicable now that the class's code has been moved into a migration script?

hixi: no, removing :-)

PR #505, opened 2016-05-24
Feature/#422 dedicated transport_l layer for air traffic

### LINKED ISSUE #422: create new table "transport_l"
Move `taxiway`, `runway`, `apron` and `aeroway` from `misc_l` to there

das-g:
closes #422

### Reviewed by
- [x] @hixi
- [x] @sfkeller (for the SQL)
- [ ] @beneditatan

You might want to review 7ff86a4 and 3677888 separately, as GitHub's overall diff view doesn't recognize that `*_air_traffic.sql` is the same (moved and modified) file.


das-g:
@hixi I've made some changes since you reviewed, thus have removed your checkmark above. Please re-review :smile:

tests/conversion/converters/bootstrap_test.py
```diff
@@ -102,36 +102,37 @@ def test_scripts_are_executed_in_correct_order(self, *args, **kwargs):
...
-                'sql/filter/create_view/021_view_traffic_p.sql',
-                'sql/filter/create_view/022_view_transport_a.sql',
-                'sql/filter/create_view/023_view_transport_p.sql',
-                'sql/filter/create_view/024_view_utility_a.sql',
-                'sql/filter/create_view/025_view_utility_p.sql',
-                'sql/filter/create_view/026_view_utility_l.sql',
-                'sql/filter/create_view/027_view_water_a.sql',
-                'sql/filter/create_view/028_view_water_p.sql',
-                'sql/filter/create_view/029_view_water_l.sql',
+                'sql/filter/create_view/0110_view_address.sql',
```
hixi: are we in need of four leading numbers?

das-g: As I had to renumber anyway, I though I'd
* group the layers
* allow for insertions within the groups

as there are 19 groups, the groups require two decimal digits. I used another two decimal digits for the numbering within the group, so that I can assign the more significant of them (the "ten"s) sequentially while still allowing for later insertions in-between using the last (and least significant) digit.

The question is, though: Do we need a decimal prefix for the view-creation SQL scripts at all? They should all be independent of each other, so we could just create them in lexicographic order of their un-prefixed script names. (We already almost do, there are just some few swaps within the "groups".)

hixi: I think lexicographic is good enough, no need for numbering, maybe for consistency only starting at 001 to 019 or so ;-)

das-g: Let's keep it simple: Now without any decimal prefix: 5ebdeb7

Issue #506, opened 2016-05-24
Remove Countries


hixi:
Remove the country app and replace migrations (squashing) in excerptexport.

Issue #502, opened 2016-05-23
Use timestamps on Export


hixi:
Use timestamps to designate process-start, process-end and process-finish time on Export

PR #507, opened 2016-05-24
Feature/remove obsolete code

hixi:
Reviewed by:
- [x] @das-g 

While working on #502 I found that we have some code around, we're not using. I cleaned this up as far as I was able to tell, with one exception: removing the countries altogether (they live on the excerpts now), but they need to be available for migrations, which is a bigger task #506, so not included here.


osmaxx/api_client/API_client.py
```diff
@@ -15,7 +15,6 @@ class RESTApiClient:
 
     def __init__(self, service_base):
         self.service_base = service_base
-        self.client = requests.session()
```
das-g: As long as we're not using this session, we can indeed delete this code. But maybe we _should_ use a [session](http://docs.python-requests.org/en/master/api/#request-sessions)? We don't need cookie persistence, but connection-pooling might be good.

And it seems that `requests.Session` already can do what we implemented ourselves with `_data_dict()` and `_default_headers()`: http://docs.python-requests.org/en/master/api/#requests.Session.headers

osmaxx/api_client/conversion_api_client.py
```diff
@@ -19,7 +14,6 @@
 USERNAME = settings.OSMAXX.get('CONVERSION_SERVICE_USERNAME')
 PASSWORD = settings.OSMAXX.get('CONVERSION_SERVICE_PASSWORD')
 
-OLD_CONVERSION_JOB_URL = '/jobs/'  # TODO: remove
```
das-g: :+1:

osmaxx/api_client/conversion_api_client.py
```diff
@@ -79,46 +73,6 @@ def _get_result_file_url(self, job_id):
...
-        output_file = OutputFile.objects.create(
-            mime_type='application/zip',
-            file_extension='zip',
-            content_type=download_file_dict['format'],
-            extraction_order=extraction_order,
-        )
-        file_name = str(output_file.public_identifier) + '.zip'
-        output_file.file = get_default_private_storage().save(file_name, ContentFile(result_response.content))
-        output_file.save()
-
```
das-g: :+1:

osmaxx/excerptexport/models/output_file.py
```diff
@@ -50,5 +50,7 @@ def get_filename_display(self):
         return ''
 
     def get_absolute_url(self):
+        if self.file:
+            return self.file.url
         from django.core.urlresolvers import reverse
-        return reverse('excerptexport:download', kwargs=dict(uuid=self.public_identifier))
+        return reverse('excerptexport:export_detail', kwargs={'id': self.export.extraction_order.excerpt.id})
```
das-g: I feel like the decision whether to use the raw download URL or the URL of a HTML view doesn't belong in this method and should instead be the concern of the caller(s).

Or maybe this function just needs to be renamed to account for this `get_absolute_url`-atypical behavior.

osmaxx/utilities/shortcuts.py
```diff
@@ -4,45 +4,11 @@
...
-    klass may be a Model, Manager, or QuerySet object. All other passed
-    arguments and keyword arguments are used in the filter() query.
-    """
-    queryset = _get_queryset(klass)
-    object_list = list(queryset.filter(*args, **kwargs))
-    if not object_list:
-        return None
-    return object_list
-
-
```
das-g: good to get rid of these two old fellas :+1:

tests/conversion/model_test.py
```diff
@@ -24,7 +24,7 @@ def test_job_get_download_url_returns_path_when_file_done(finished_conversion_jo
 @pytest.mark.django_db()
 def test_job_removes_file_when_deleted(finished_conversion_job):
     assert finished_conversion_job.status == FINISHED
-    file_path = finished_conversion_job.resulting_file.name
+    file_path = finished_conversion_job.resulting_file.path
```
das-g: was this test subtly broken (but still passing) before?

tests/urls.py
```diff
@@ -9,4 +14,6 @@
     url(r'^api/', include('osmaxx.excerptexport.rest_api.urls', namespace='excerptexport_api')),
     url(r'^job_progress/', include('osmaxx.job_progress.urls', namespace='job_progress')),
     url(r'^pages/', include('osmaxx.core.urls', namespace='pages')),
+    # we're serving through nginx in production, but we need this here so we can test more easily!
+    url(r'^%s(?P<path>.*)$' % re.escape(settings.MEDIA_URL.lstrip('/')), serve, kwargs={'document_root': settings.MEDIA_ROOT}),
```
das-g: We usually use `'...'.format(...)` for string substitution, rather than the `%` operator. Consider using it here, too.

osmaxx/api_client/API_client.py
hixi: I realized that, and AFAIK that line stems from an idea where I implemented it this way - but forgot about it again, and didn't remember while reviewing.

I created the issue #508 for this, since this doesn't belong in here.

tests/conversion/model_test.py
hixi: - `name` is not the path
- using a real world scenario now (with updated settings in conftests, which defines a `MEDIA_ROOT`), this failed
- now the test tests what it actually meant to test ;-)

tests/urls.py
hixi: I'm merely using the same syntax as in the django documentation, I think that is OK and more easy to find that way, when searching for [it](https://docs.djangoproject.com/en/1.9/ref/contrib/staticfiles/#static-file-development-view) on the web

osmaxx/excerptexport/models/output_file.py
hixi: good idea :-)

osmaxx/api_client/API_client.py
das-g: :+1:

tests/urls.py
das-g: Huh? I don't see the `%` operator being used on that documentation page.

hixi: Oh, because it isn't there ;-)

It is here:
https://github.com/django/django/blob/master/django/conf/urls/static.py

hixi:
@das-g I made some changes, so I took the liberty of removing your check-mark above.

PR #513, opened 2016-05-26
Feature/#502 use timestamps on export

### LINKED ISSUE #468: Show file size besides download links
Requested by @geoSLG for "My Orders" page.

Probably makes sense for other pages where we display export download links, too.

### LINKED ISSUE #469: Show date besides download links
Requested by @geoSLG for "My Orders" page.

Probably makes sense for other pages where we display export download links, too.

I (@das-g) think the date & time of the OSM data that went into the export would be most useful, but for now, the order date will probably do.

### LINKED ISSUE #502: Use timestamps on Export

hixi:
Does abit more than requested on the issue:
- uses timestamps
- displays them
- updates them when a file is downloaded

Reviewed by:

- [x] @das-g 

Closes #502 
Closes #468
Closes #469


osmaxx/excerptexport/models/export.py
```diff
@@ -1,21 +1,33 @@
...
 
 
-class Export(models.Model):
+class TimeStampModelMixin(models.Model):
+    created = models.DateTimeField(_('created on'), default=datetime.now, blank=True, editable=False)
+    updated = models.DateTimeField(_('updated on'), default=None, blank=True, editable=False, null=True)
+
+    def save(self, *args, **kwargs):
+        now = datetime.now()
+        if not self.id:
```
das-g: 0 is falsy!

osmaxx/excerptexport/templates/excerptexport/partials/export_display.html
```diff
@@ -31,5 +30,11 @@
                 </div>
             </div>
         </div>
+    {% elif export.status == "finished" %}{# FIXME: finished is hardcoded, but as long as we don't have the STATUS on the model, we can't access it here! #}
```
das-g: :bulb: `import` the stati in [the model class](https://github.com/geometalab/osmaxx-frontend/pull/513/files#diff-e52c21d719a322835dc4db89ea595076R30) to make them accessible on the model instances.

osmaxx/excerptexport/static/excerptexport/scripts/maps/excerpt_viewer.js
```diff
@@ -1,8 +1,22 @@
-window.ExcerptViewer = function(mapElementID, excerptApiUrl) {
+window.ExcerptViewer = function(mapElementID, excerptApiUrl, disableZoom) {
```
das-g: [No positional boolean arguments](http://www.informit.com/articles/article.aspx?p=1392524), please.

osmaxx/excerptexport/templates/excerptexport/export_list.html
```diff
@@ -7,26 +7,25 @@
...
+                            <h3 class="panel-title">{{ excerpt.name }}</h3>
                         </div>
                         <div class="panel-body">
                             <div class="col-lg-6">
-                                {% for export in export_group.list|slice:":5" %}
+                                {% for export in exports|slice:":5" %}
                                     {% include 'excerptexport/partials/export_display.html' %}
                                 {% endfor %}
-                                {% if export_group.list|length > 5 %}<a href="{% url 'excerptexport:export_detail' id=export_group.grouper.id %}">{% trans 'see all your exports of this excerpt.' %} ({{ export_group.list|length }})</a>{% endif %}
+                                {% if exports|length > 5 %}<a href="{% url 'excerptexport:export_detail' id=excerpt.id %}">{% trans 'see all your exports of this excerpt.' %} ({{ exports|length }})</a>{% endif %}
```
das-g: :+1: much easier to grasp, like this!

osmaxx/excerptexport/views.py
```diff
@@ -69,31 +73,91 @@ def get_object(self, queryset=None):
...
+            )
+        return self._excerpt_ids
+
+    @property
+    def status_choices(self):
+        allowed_choices = [
+            statuses.FINISHED,
+            statuses.FAILED,
+        ]
+        return [choice for choice in Export.STATUS_CHOICES if choice[0] in allowed_choices]
```
das-g: The name "allowed choices" here is misleading:

- This list holds just the statuses, **not yet the choices** (those are returned by this method)
- "Allowed" in what sense, exactly? Surely our Exports can have other statuses than just these two.

Should these be `filterable_statuses` or are they also used for something else?

osmaxx/excerptexport/views.py
```diff
@@ -69,31 +73,91 @@ def get_object(self, queryset=None):
...
+
+    def get_user_exports(self):
+        return self._filter_exports(
+            Export.objects.filter(extraction_order__orderer=self.request.user)
+        ).order_by('-updated', '-finished')
+
+    def _filter_exports(self, query):
+        allowed_status_filters = [
+            status[0] for status in self.status_choices
+        ]
```
das-g: maybe easier to comprehend (sic!) as
```python
        allowed_status_filters = [
            status for status, _ in self.status_choices
        ]
```
i.e. unpacking instead of explicit indexing.

osmaxx/excerptexport/views.py
```diff
@@ -69,31 +73,91 @@ def get_object(self, queryset=None):
...
+        status_filter = self.request.GET.get('status', None)
+        if status_filter and status_filter in allowed_status_filters:
```
das-g:
```python
if status_filter in allowed_status_filters:
```
should suffice here. `None` can be used on the left side of containment tests just fine. (`None in [None, 2, 'three']` is `True` while `None in [1, 2, 'three']` is `False`.)

osmaxx/excerptexport/views.py
```diff
@@ -69,31 +73,91 @@ def get_object(self, queryset=None):
...
+    paginate_by = 5
+
+    def get_context_data(self, **kwargs):
+        context_data = super().get_context_data(**kwargs)
+        context_data.update(self._get_extra_context_data())
+        context_data['excerpt_list_with_exports'] = OrderedDict(
+            (
+                excerpt,
+                self.get_user_exports().filter(extraction_order__excerpt__pk=excerpt.id)
+                .select_related('extraction_order', 'extraction_order__excerpt', 'output_file'))
```
das-g: Put the second closing `)` on the next line. Either together with the `for ...` or on a line of its own.

das-g: Could `.filter(extraction_order__excerpt__pk=excerpt.id)` be written shorter as `.filter(extraction_order__excerpt=excerpt)`, or would that give different results?

das-g: consider extracting `self.get_user_exports().filter(extraction_order__excerpt__pk=excerpt.id).select_related('extraction_order', 'extraction_order__excerpt', 'output_file')` to a function, so that the pairs (2-tuples) inside the generator expression become more apparent for the OrderedDict initialization without having to spell them out as `tuple(..., ...)`.

tests/excerptexport/models/test_export_model.py
```diff
@@ -0,0 +1,31 @@
...
+
+@pytest.mark.django_db
+def test_created_export_has_correct_time_stamp(export):
+    now = datetime.now()
+    margin = timedelta(minutes=1)
+    assert export.finished is None
+    assert export.updated is not None
+    assert export.created is not None
+    assert export.created == export.updated
+    assert (now - margin) < export.created < (now + margin)
```
das-g: Could also be written as
```python
assert abs(export.created - now) < margin
```
but explicit bounds are fine, too.

Could this be tested without margin by mocking out `datetime.now()`?

tests/excerptexport/models/test_export_model.py
```diff
@@ -0,0 +1,31 @@
+from datetime import timedelta
+
+import pytest
+
+from django.utils.datetime_safe import datetime
+
+
+@pytest.mark.django_db
+def test_created_export_has_correct_time_stamp(export):
```
das-g: As `Export.objects.create` is the action being tested here, consider moving that from the fixture to the test body. To avoid duplication between the test and the fixture, this can be done by explicitly invoking the fixture's function. (Import it locally to avoid namespace pollution.)

tests/excerptexport/models/test_export_model.py
```diff
@@ -0,0 +1,31 @@
...
+    assert export.created == export.updated
+    assert (now - margin) < export.created < (now + margin)
+
+
+@pytest.mark.django_db
+def test_saved_again_export_has_correct_time_stamp(export):
+    export.save()
+    assert export.updated is not None
+    assert export.created is not None
+    assert export.created != export.updated
```
das-g: Mock `datetime.now` with a known return value during `export.save()` above to guarantee this assertion passes even for magic faster-than-timestamp-resolution computers.

osmaxx/excerptexport/models/export.py
```diff
@@ -1,21 +1,33 @@
...
-INITIAL_CHOICE = (INITIAL, _('initial'))
-STATUS_CHOICES = (INITIAL_CHOICE,) + statuses.STATUS_CHOICES
-
 logger = logging.getLogger(__name__)
 
 
-class Export(models.Model):
+class TimeStampModelMixin(models.Model):
+    created = models.DateTimeField(_('created on'), default=datetime.now, blank=True, editable=False)
+    updated = models.DateTimeField(_('updated on'), default=None, blank=True, editable=False, null=True)
```
das-g: As these are datetimes, not bare dates, use "at" rather than "on" for the display name. You might want to rename the fields themselves to `created_at` and `updated_at`, too, as the current names seem to imply a boolean type.

osmaxx/excerptexport/models/export.py
```diff
@@ -27,11 +39,16 @@ class Export(models.Model):
...
+    INITIAL = 'initial'
+    INITIAL_CHOICE = (INITIAL, _('initial'))
+    STATUS_CHOICES = (INITIAL_CHOICE,) + statuses.STATUS_CHOICES
+
     extraction_order = models.ForeignKey('excerptexport.ExtractionOrder', related_name='exports',
                                          verbose_name=_('extraction order'))
     file_format = models.CharField(choices=FORMAT_CHOICES, verbose_name=_('file format / data format'), max_length=10)
     conversion_service_job_id = models.IntegerField(verbose_name=_('conversion service job ID'), null=True)
     status = models.CharField(_('job status'), choices=STATUS_CHOICES, default=INITIAL, max_length=20)
+    finished = models.DateTimeField(_('finished on'), default=None, blank=True, editable=False, null=True)
```
das-g: As these is a datetime, not a bare date, use "at" rather than "on" for the display name. You might want to rename the field itselve to `finished_at`, too, as the current name seems to imply a boolean type.

tests/excerptexport/test_downloads.py
```diff
@@ -1,7 +1,26 @@
...
+
+def test_successful_file_attaching_changes_export_finished_timestamp(mocker, output_file_content, export):
+    margin = timedelta(minutes=1)
+    now = datetime.now()
+    mocker.patch.object(
+        ConversionApiClient, 'get_result_file',
+        side_effect=[ContentFile(output_file_content)],
+    )
+    assert export.finished is None
+    assert export.status is export.INITIAL
```
das-g: Why check `export.status` before fetching if you don't check it afterwards, too?

osmaxx/excerptexport/static/excerptexport/scripts/maps/excerpt_viewer.js
hixi: I don't think that he knew about JavaScript ;-)

Translating to JavaScript it would give us:
```javascript
function(mapElementID, excerptApiUrl, options) {
    if (!options) options = {};
    if (!options.disableZoom) options.disableZoom = False;
...
}
``

I don't think this makes this in any way clearer.

hixi: after a short discussion: using a function that can be called should solve this issue quite neatly, doing that now.

osmaxx/excerptexport/views.py
hixi: :+1:

osmaxx/excerptexport/views.py
hixi: Should probably be extracted to a property, since it is used also in `_filter_exports`.

osmaxx/excerptexport/views.py
hixi: You're right, forgot about that. :+1:

osmaxx/excerptexport/views.py
hixi: :+1: for extracting to function and reducing complexity with leaving out the `__pk` part.

das-g: > with leaving out the `__pk` part

also the `.id` part

osmaxx/excerptexport/templates/excerptexport/partials/export_display.html
hixi: :+1:

tests/excerptexport/models/test_export_model.py
hixi: Probably can, but would require more test-code and make the test a bit harder.

tests/excerptexport/test_downloads.py
hixi: good argument :smiling_imp:

tests/excerptexport/models/test_export_model.py
hixi: Importing is not a good idea, as it would need a request object to be constructed, which is more finicky than having the same code copied over again.

das-g:
:+1:

PR #514, opened 2016-05-27
Feature/make integration tests work again

hixi:
Reviewed by:
- [x] @das-g 

Closes no issue.


docker-compose-dev.yml
```diff
@@ -70,7 +70,16 @@ services:
...
+    command: >
+      sh -c "
+      mkdir -p /var/data/osm-planet/pbf &&
+      wget -O /tmp/monaco-latest.osm.pbf http://download.geofabrik.de/europe/monaco-latest.osm.pbf &&
+      wget -O /tmp/switzerland-latest.osm.pbf http://download.geofabrik.de/europe/switzerland-latest.osm.pbf &&
+      osmconvert /tmp/monaco-latest.osm.pbf -o=/tmp/monaco-latest.osm &&
+      osmconvert /tmp/switzerland-latest.osm.pbf -o=/tmp/switzerland-latest.osm &&
+      osmconvert /tmp/monaco-latest.osm /tmp/switzerland-latest.osm -o=/var/data/osm-planet/pbf/planet-latest.osm.pbf &&
+      osm2pgsql -c --number-processes 8 -H database -U gis -d gis --slim -C 2000 --prefix osm --hstore --latlon --style /root/styles/terminal.style --tag-transform-script /root/styles/tag_transform_style.lua /var/data/osm-planet/pbf/planet-latest.osm.pbf
+      "
```
das-g: already much nicer, but why not a separate file for this script?

hixi: firstly, because it would clutter the workspace, which already has many files in the root directory.

secondly, because it would need to be injected into the container, linking another volume in there.

thirdly, because it is only for development, and there really is no need to have it anywhere else, other than in the `docker-compose-dev.yml`.

tests/selenium_tests/new_excerpt.py
```diff
@@ -11,3 +11,21 @@ def new_excerpt(driver, base_url):
...
     #                           | South = 43.7289719167851 | West = 7.41568565368652)
     geojson_field.send_keys('{"type":"Polygon","coordinates":[[[7.423303127288818,43.7380498124601],[7.418947219848633,43.73685604310849],[7.418625354766846,43.73373972784547],[7.420299053192139,43.73046819754524],[7.425191402435303,43.72959989409132],[7.428925037384032,43.72961539962052],[7.432680130004883,43.73491805517113],[7.433323860168456,43.737662227617484],[7.428624629974364,43.74053029604388],[7.422015666961669,43.74063881485443],[7.42100715637207,43.74180150405227],[7.418088912963867,43.741817006422366],[7.423303127288818,43.7380498124601]]]}')
+
+
+def new_excerpt_through_admin(selenium, base_url, excerpt_name="existing_excerpt"):
+    selenium.get('{0}/admin/excerptexport/excerpt/'.format(base_url))
+    try:
+        selenium.find_element_by_xpath("//a[contains(text(),'{}')]".format(excerpt_name))
+        return
+    except:
```
das-g: consider only swallowing a more specific error

tests/selenium_tests/new_excerpt.py
```diff
@@ -11,3 +11,21 @@ def new_excerpt(driver, base_url):
...
+        pass
+    selenium.get('{0}/admin/excerptexport/excerpt/add/'.format(base_url))
+    name_field = selenium.find_element_by_id('id_name')
+    name_field.send_keys(excerpt_name)
+    geojson_field = selenium.find_element_by_id('id_bounding_geometry')
+    geojson_field.send_keys("SRID=4326;MULTIPOLYGON (((7.4233031272888184 43.7380498124600976, 7.4189472198486328 43.7368560431084887, 7.4186253547668457 43.7337397278454674, 7.4202990531921387 43.7304681975452425, 7.4251914024353027 43.7295998940913222, 7.4289250373840323 43.7296153996205206, 7.4326801300048828 43.7349180551711285, 7.4333238601684561 43.7376622276174842, 7.4286246299743643 43.7405302960438789, 7.4220156669616690 43.7406388148544281, 7.4210071563720703 43.7418015040522690, 7.4180889129638672 43.7418170064223659, 7.4233031272888184 43.7380498124600976)))")
+    from selenium.webdriver.support.ui import Select
```
das-g: Is there a reason to import this only locally?

tests/selenium_tests/test_new_excerpt.py
```diff
@@ -26,10 +26,13 @@ def test_new_excerpt(base_url, login, file_name, file_format, selenium, reload_u
     create.send_keys(Keys.RETURN)
 
     # wait until download link appears
-    selenium.find_element_by_link_text('↻ Reload')
-    element = reload_until_condition(selenium.find_element_by_class_name, "form-control")
+    selenium.find_element_by_xpath("/html/body/div/div/div[2]/div[2]/div[1]/div/div[1]/h3")
+    first_a = "/html/body/div/div/div[2]/div[2]/div[1]/div/div[2]/div[1]/div[1]/div[2]/div/div[1]/p/a"
```
das-g: `find_element_by_link_text` was easier to read and matches more closely what a user would do. (They usually don't care about the DOM as long as the page looks good and they can find what they're looking for.) What's the reason to switch to `find_element_by_xpath`?

I guess it isn't arbitrary that these two XPaths have the common prefix `/html/body/div/div/div[2]/div[2]/div[1]/div/`. Should that prefix be extracted to a constant or local variable, so that it's name can give the reader some hint on what we're looking for in the DOM here?

tests/selenium_tests/test_new_excerpt.py
```diff
@@ -26,10 +26,13 @@ def test_new_excerpt(base_url, login, file_name, file_format, selenium, reload_u
...
+    element = reload_until_condition(selenium.find_element_by_xpath, first_a)
 
     # check if the download link is a valid link
-    url = element.text
+    url = element.get_attribute('href')
+    if not url.startswith('http'):
+        url += 'http://{}{}'.format(base_url, url)
```
das-g: Using [`urllib.parse.urljoin`](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin) would be more reliable. (Handle's absolute, schema-relative, network-location-relative and path-relative URLs just like browsers should do.)

```python
    from urllib.parse import urljoin
    url = urljoin(base_url, element.get_attribute('href'))
```

tests/selenium_tests/test_select_country.py
```diff
@@ -29,10 +29,13 @@ def test_select_country(base_url, login, file_format, selenium, reload_until_con
     create.send_keys(Keys.RETURN)
 
     # wait until the download link appears
-    selenium.find_element_by_link_text('↻ Reload')
-    element = reload_until_condition(selenium.find_element_by_class_name, "form-control")
+    selenium.find_element_by_xpath("/html/body/div/div/div[2]/div[2]/div[1]/div/div[1]/h3")
+    first_a = "/html/body/div/div/div[2]/div[2]/div[1]/div/div[2]/div[1]/div[1]/div[2]/div/div[1]/p/a"
+    element = reload_until_condition(selenium.find_element_by_xpath, first_a)
```
das-g: [dito](https://github.com/geometalab/osmaxx-frontend/pull/514/files#r65045557)

tests/selenium_tests/test_select_country.py
```diff
@@ -29,10 +29,13 @@ def test_select_country(base_url, login, file_format, selenium, reload_until_con
...
 
     # check if the download link is a valid link
-    url = element.text
+    url = element.get_attribute('href')
+    if not url.startswith('http'):
+        url += 'http://{}{}'.format(base_url, url)
```
das-g: [dito](https://github.com/geometalab/osmaxx-frontend/pull/514/files#r65048361)

tests/selenium_tests/test_new_excerpt.py
hixi: The reason was simply that we need to get the first of the div's and select it in there. I haven't tried the same with `find_element_by_link_text` because the tests take such a very long time to run, but will take a look if it might be possible to do so.

tests/selenium_tests/test_new_excerpt.py
hixi: and looks nicer :-)

tests/selenium_tests/new_excerpt.py
hixi: :+1:

tests/selenium_tests/test_new_excerpt.py
hixi: oh, this doesn't work at all, since we go to a page where we have link texts, but the first one isn't the one we're waiting on, it is the first one that is already finished, which is not what we want.

Extracting to a constant is a good idea.

tests/selenium_tests/new_excerpt.py
hixi: no.

tests/selenium_tests/conftest.py
```diff
@@ -60,3 +60,6 @@ def wait_for_page_load(browser, timeout=30):
     WebDriverWait(browser, timeout).until(
         staleness_of(old_page)
     )
+
+
+first_panel_on_excerpts_export_overview = '/html/body/div/div/div[2]/div[2]/div[1]/div/'
```
das-g: As selenium also allows to put objects representing actual DOM elements into variables, indicate in the variable name here that this is an XPath path.

PR #518, opened 2016-05-28
Feature/splitted large countries

### LINKED ISSUE #517: Reduce selection to subpregions
For Germany, USA, Russian Federation, France and The Netherlands only allow selection of their respective subregions.

hixi:
Reviewed by:
- [x] @das-g 

Closes #517


osmaxx/excerptexport/migrations/0039_update_countries_20160528_2049.py
```diff
@@ -0,0 +1,45 @@
...
+                is_public=True,
+                name=name,
+                excerpt_type='country',
+            )
+            update_excerpt(excerpt, polyfile_path)
+        else:
+            for excerpt in existing_excerpts:
+                update_excerpt(excerpt, polyfile_path)
+    # Remove old countries. Yes, this deletes the existing exports of those countries as well!
+    Excerpt.objects.filter(excerpt_type='country').exclude(name__in=done_countries).delete()
```
das-g: ok :boom:

osmaxx/utils/polyfile_helpers.py
```diff
@@ -67,3 +67,23 @@ def parse_poly(lines):
...
+def get_polyfile_names_to_file_mapping():
+    """
+    generates names of all polyfiles inside the POLYFILE_LOCATION path.
+
+    Returns: a mapping of polyfile names to polyfile path
+    """
+    from osmaxx.excerptexport._settings import POLYFILE_LOCATION
+    polyfile_name_to_file_path_mapping = {}
+    for root, dirs, files in os.walk(POLYFILE_LOCATION):
+        subfolder_name = root.replace(POLYFILE_LOCATION, '').replace('/', '')
```
das-g: What if `POLYFILE_LOCATION` re-occurs within `root`? Maybe use
```python
subfolder_name = root[len(POLYFILE_LOCATION):].replace('/', '')
```

osmaxx/utils/polyfile_helpers.py
```diff
@@ -67,3 +67,23 @@ def parse_poly(lines):
...
+        for possible_polyfile in files:
+            name_parts = [] if subfolder_name == '' else [subfolder_name]
+            if _is_polyfile(possible_polyfile):
```
das-g: `name_parts` is only used inside the `if` clause, so maybe define it there.

osmaxx/utils/polyfile_helpers.py
hixi: not very likely, since `POLYFILE_LOCATION` should be an absolute path, but I like the explicitness of your expression :-)

PR #520, opened 2016-05-30
Feature/#423 nonop way classification

### LINKED ISSUE #423: add classification to `nonop` features
To be decided: how should the column be named

Add a new column, ie. `sub_type` which holds which sepcific type it is

das-g:
closes #423

we only handle
* [`<main tag key>=<status>`+`<status>=<main tag value>`, e.g. `highway=construction`+`construction=tertiary`](http://wiki.openstreetmap.org/wiki/Comparison_of_life_cycle_concepts#.3Ckey.3E_.3D_.3Cstatus.3E_.2B_.3Cstatus.3E_.3D_.3Cvalue.3E)
* `<main tag key>=<status>`, e.g. `highway=construction` (using the fallback `sub_type`s in this case)

[There are more lifecycle tagging schemes](http://wiki.openstreetmap.org/wiki/Comparison_of_life_cycle_concepts), most notable [lifecycle prefix](http://wiki.openstreetmap.org/wiki/Lifecycle_prefix) (`<status>:<main tag key>=<main tag value>`, e.g. `construction:highway=tertiary`) that we didn't handle before this PR and don't handle with it (yet), either.

### Reviewed by
- [x] @hixi
- [ ] @sfkeller

### DIFF
--- a/osmaxx/conversion/converters/converter_gis/bootstrap/sql/filter/nonop/005_lifecycle_view.sql
+++ b/osmaxx/conversion/converters/converter_gis/bootstrap/sql/filter/nonop/005_lifecycle_view.sql
@@ -3,16 +3,16 @@ CREATE OR REPLACE VIEW osmaxx.lifecycle_l AS
     osm_id,
     CASE
       WHEN highway = 'planned' THEN tags -> 'planned'
-      WHEN highway = 'disused' THEN tags -> 'disused'
       WHEN highway = 'construction' THEN tags -> 'construction'
+      WHEN highway = 'disused' THEN tags -> 'disused'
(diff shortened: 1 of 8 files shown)


osmaxx/conversion/converters/converter_gis/bootstrap/sql/filter/nonop/005_lifecycle_view.sql
```diff
@@ -0,0 +1,34 @@
+CREATE OR REPLACE VIEW osmaxx.lifecycle_l AS
```
hixi: very nice having an "abstraction" for this :+1:

PR #547, opened 2016-06-07
Feature/#544 remove uuid from filename

### LINKED ISSUE #544: remove uuid from outputfile filename
Remove the UUID from the filename, it already exists in the folder as security measure.

hixi:
Reviewed by:
- [x] @das-g 

Closes #544


osmaxx/excerptexport/models/output_file.py
```diff
@@ -38,6 +36,13 @@ def download_file_name(self):
     def content_type(self):
         return self.export.file_format
 
+    @property
+    def file_extension(self):
+        if bool(self.file):
+            _discarded, file_extension = os.path.splitext(self.file)
```
das-g: I guess you're using this name, because just `_` would shadow the `_` from
```python
from django.utils.translation import ugettext_lazy as _
```
?

hixi: correct, I didn't want to shadow that name.

das-g:
:+1:

PR #555, opened 2016-06-08
Feature/#554 remove files from shared worker storage

### LINKED ISSUE #554: remove files from shared worker storage when adding as outputfile
now that the storage from the worker is connected to the frontend, remove the corresponding file from the shared storage whenever it is attached to an output file.

Better even, use move the file instead of copying:
https://docs.djangoproject.com/en/1.10/topics/files/

hixi:
Reviewed by:
- [x] @das-g 

Closes #554

### DIFF
--- a/osmaxx/excerptexport/models/__init__.py
+++ b/osmaxx/excerptexport/models/__init__.py
@@ -1,7 +1,7 @@
 from .excerpt import Excerpt
 from .export import Export
 from .extraction_order import ExtractionOrder, ExtractionOrderState
-from .output_file import OutputFile, uuid_directory_path
+from .output_file import OutputFile
 
 __all__ = [
     "Excerpt",
@@ -9,5 +9,4 @@ __all__ = [
     "ExtractionOrder",
     "ExtractionOrderState",
     "OutputFile",
-    "uuid_directory_path",
 ]
--- a/osmaxx/excerptexport/models/__init__.py
+++ b/osmaxx/excerptexport/models/__init__.py
@@ -1,5 +1,13 @@
-# flake8: noqa
 from .excerpt import Excerpt
 from .export import Export
 from .extraction_order import ExtractionOrder, ExtractionOrderState
-from .output_file import OutputFile
+from .output_file import OutputFile, uuid_directory_path
+
+__all__ = [
+    "Excerpt",
+    "Export",
+    "ExtractionOrder",
+    "ExtractionOrderState",
+    "OutputFile",
+    "uuid_directory_path",
+]
--- a/osmaxx/excerptexport/models/export.py
+++ b/osmaxx/excerptexport/models/export.py
@@ -113,7 +113,8 @@ class Export(TimeStampModelMixin, models.Model):
 
     def _fetch_result_file(self):
         from osmaxx.api_client import ConversionApiClient
-        from . import OutputFile, uuid_directory_path
+        from . import OutputFile
+        from osmaxx.excerptexport.models.output_file import uuid_directory_path
         api_client = ConversionApiClient()
         file_path = api_client.get_result_file_path(self.conversion_service_job_id)
         now = timezone.now()
--- a/osmaxx/excerptexport/models/export.py
+++ b/osmaxx/excerptexport/models/export.py
@@ -1,5 +1,7 @@
 import logging
+import os
 
+from django.conf import settings
 from django.db import models
 from django.utils import timezone
 from django.utils.translation import ugettext_lazy as _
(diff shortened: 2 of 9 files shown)


osmaxx/excerptexport/models/__init__.py
```diff
@@ -1,5 +1,13 @@
-# flake8: noqa
 from .excerpt import Excerpt
 from .export import Export
 from .extraction_order import ExtractionOrder, ExtractionOrderState
-from .output_file import OutputFile
+from .output_file import OutputFile, uuid_directory_path
```
das-g: consider leaving out `uuid_directory_path` here, as it isn't a model class

tests/excerptexport/test_downloads.py
```diff
@@ -12,14 +10,29 @@ def test_file_download(authenticated_client, user, output_file_with_file, output
...
+
+
+def test_successful_file_attaching_removes_original_file(mocker, some_fake_zip_file, export):
+    mocker.patch.object(
+        ConversionApiClient, 'get_result_file_path',
+        side_effect=[some_fake_zip_file.name],
+    )
+    assert os.path.exists(some_fake_zip_file.name)
+    export._fetch_result_file()
+    assert bool(export.output_file.file)
```
das-g: AFAIK, `assert` (like other control structures expecting a boolean context, such as `if` and `while`) already implicitly casts to `bool`. Is the explicit cast here for readability?

tests/excerptexport/test_downloads.py
```diff
@@ -12,14 +10,29 @@ def test_file_download(authenticated_client, user, output_file_with_file, output
...
+    assert not os.path.exists(some_fake_zip_file.name)
+    assert os.path.exists(export.output_file.file.path)
+    from osmaxx.excerptexport.models import uuid_directory_path
+    from django.conf import settings
+    assert export.output_file.file.path == os.path.join(settings.MEDIA_ROOT, uuid_directory_path(export.output_file, some_fake_zip_file.name))
```
das-g: Will the (moved) temporary file be deleted after the test?

hixi: oops, I don't think so. Good catch, thanks :-)

osmaxx/excerptexport/models/__init__.py
hixi: :+1:

tests/excerptexport/test_downloads.py
hixi: now I remember, yes, boolean cast is not neccessary; but it is not `None`, so that comparison fails, even if the field specifies `null=True` - this lead to confusion in the past and I remembered wrongly. Thanks for pointing this out.

web_frontend/config/settings/local.py
```diff
@@ -9,7 +9,18 @@
...
+
+
+class glob_list(list):  # noqa
+    def __contains__(self, key):
+        for elt in self:
+            if fnmatch(key, elt):
+                return True
+        return False
+
+INTERNAL_IPS = glob_list(env.tuple('DJANGO_INTERNAL_IPS', default=('127.0.0.1',)))
```
das-g: How about something like this? (untested so far)
```python
class Subnet():
    def __init__(self, ip_as_int, mask_as_int):
        self._ip = ip_as_int
        self._mask = mask_as_int
    
    @classmethod
    def from_str(cls, s):
        m = re.match('^(\d{1,3}).(\d{1,3}).(\d{1,3}).(\d{1,3})/(\d+)$', s)
        ip_as_int = int.from_bytes((int(x) for x in m.groups()[:4]), byteorder='big')
        mask_as_int = (0xFFFFFFFF<<(32 - m.groups()[5])) & 0xFFFFFFFF
        return cls(ip_as_int, mask_as_int)
        
    def __contains__(self, key):
        m = re.match('^(\d{1,3}).(\d{1,3}).(\d{1,3}).(\d{1,3})$', key)
        ip_as_int = int.from_bytes((int(x) for x in m.groups()), byteorder='big')
        return ip_as_int & self.mask_as_int == self.ip_as_int
        
INTERNAL_IPS = chain(Subnet(s) for s in env.tuple('DJANGO_INTERNAL_IPS', default=('127.0.0.1',)))
```

hixi: looks interesting :+1:

das-g: `itertools.chain` requires `self.__next__()` to work.

My approach seems to work with `collections.ChainMap` instead of `itertools.chain`, though. :-)

osmaxx/excerptexport/models/export.py
```diff
@@ -113,7 +113,8 @@ def _get_job_finished_but_result_file_missing_log_message(self):
 
     def _fetch_result_file(self):
         from osmaxx.api_client import ConversionApiClient
-        from . import OutputFile, uuid_directory_path
+        from . import OutputFile
+        from osmaxx.excerptexport.models.output_file import uuid_directory_path
```
das-g: If you want to stick with a relative import for this,
```python
from .output_file import uuid_directory_path
```
should work, I think.

das-g:
:+1:

PR #571, opened 2016-06-09
Feature/#563 show download only for logged in users

### LINKED ISSUE #563: Downloads Page adaptations part 1
- [ ] display download page only for logged in users
- [ ] add dowloads as separat tab, ie. before `about`

hixi:
Reviewed by:
- [x] @das-g 

Closes #563

### DIFF
--- a/osmaxx/core/templates/osmaxx/footer.html
+++ b/osmaxx/core/templates/osmaxx/footer.html
@@ -1,1 +1,1 @@
-<p class="footer">&copy; 2015-2016 HSR Hochschule für Technik, Rapperswil | <a href="{% url 'pages:downloads' %}">Additional Downloads</a></p>
+<p class="footer">&copy; 2015-2016 HSR Hochschule für Technik, Rapperswil</p>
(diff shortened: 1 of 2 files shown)


osmaxx/core/templates/osmaxx/footer.html
```diff
@@ -1 +1 @@
-<p class="footer">&copy; 2015-2016 HSR Hochschule für Technik, Rapperswil</p>
```
das-g: That is cheating! :cop: The copyright date should _only_ be bumped when a **new creative work** is derived from the original creative work. Just keeping the application deployed and running on a server might be hard work, too, but doesn't add new creativity to the original work (unless the means to keep it running are creative and made part of the original work, e.g. by checking in new configurations or scripts).

The AFAIK only reason for giving a date (or date range) with the copyright notice is that after a fixed period (the length of which depends on the applicable legislation), the work leaves copyright protection and enters the public domain. The date helps readers to determine when that is the case. If a range is given, adding the length of the fixed period to the earlier date gives the time when the first parts of the work enter the public domain (not so useful for the reader, because they'd have to figure out what parts those are to be able to use them freely), adding it to the later date gives the time when the complete work (in its current state) will enter the public domain.

As always, IANAL & TINLA.

PR #575, opened 2016-06-14
fix #574: use addr:city=* rather than addr:place=* as city in addresses

### LINKED ISSUE #574: attribute "city" of layer "address_p" should use OSM tag "addr:city", not "addr:place"
Use OSM tag `addr:city` rather than `addr:place` for attribute `city` of layer `address_p` in all SQL scripts where that attribute is set.

---
##### original description
~~[Currently, we put the value](https://github.com/geometalab/osmaxx-frontend/blob/86cb6871ed7b02b3aa9160db0a5d079ccd1e5aee/osmaxx/conversion/converters/converter_gis/bootstrap/sql/filter/address/010_address.sql#L35) of [OSM tag `addr:place=*`](http://wiki.openstreetmap.org/wiki/Key:addr:place) into the `city` attribute of layer `address_p`.~~

~~I think if present, tags [`addr:city=*`, ` is_in:city=*`, `addr:hamlet=*` and `addr:suburb=*`](http://wiki.openstreetmap.org/wiki/Key:addr) should be used instead (if any of them present, precedence if several present TDB.)~~

~~If none of them are present, we can still fall back to `addr:place=*`. Although, we might only want to do that if `addr:street=*` is present, so that the `addr:place=*` hasn't already been used in attribute `street`.~~

das-g:
fixes #574 

Use value of OSM tag `addr:city=*` rather than `addr:place=*` as attribute `city` in layer `address_p`.

### Reviewed by
- [x] @hixi
- [ ] @sfkeller

### DIFF
--- a/osmaxx/conversion/converters/converter_gis/bootstrap/sql/filter/address/010_address.sql
+++ b/osmaxx/conversion/converters/converter_gis/bootstrap/sql/filter/address/010_address.sql
@@ -32,7 +32,7 @@ INSERT INTO osmaxx.address_p
     end  as street,
     "addr:housenumber" as housenumber,
     "addr:postcode" as postcode,
-    "addr:place" as city,
+    "addr:city" as city,
     "addr:country" as country
   FROM osm_point
 where building !='entrance' and entrance is null and ("addr:street" !='' OR "addr:housenumber"!='' OR "addr:place"!='')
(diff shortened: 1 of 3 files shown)


hixi:
:+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1:

PR #576, opened 2016-06-15
Feature/#368 delete old file from frontend server

### LINKED ISSUE #368: Delete old download files from frontend server
As an operator, I don't want export results (large-ish ZIP archives) to take up disk space forever. They should be deleted automatically somewhen to make space for new results.

As a user, want to have the result files available for a set minimum time after I'm notified that they've become available, so I can
- come and get them later, in case I can't or don't want to react to the notification right away
- send the download links to others who'll then also have enough time to go get the files if they want
- re-download recent files in case I loose or destroy my local copy, or don't have the drive I've put them on with me and need the file on another device

For this, I'd like to know how long (and/or even better, until when) a given file will at least be available.

---

### Implementation notes:
- Downloads expire after a given constant (configurable by setting) time after the user has been first notified about them (note that notification time might be considerably after creation time)
- Download expiration time: 1 month
- Each file's expiration date should be explicitly stored, so that it won't change for existing files when the setting is adapted
- Expiry date (and maybe minimum availability duration, too) should be visible on
  - [ ] the order's page, in the [email the orderer receives](https://github.com/geometalab/osmaxx/blob/dda2d72654126f33d9cc2b0d66e5f5f1f417172a/web_frontend/osmaxx/job_progress/tests.py#L175)
  - in the emails the orderer can send do others
     - [ ] [(1)](https://github.com/geometalab/osmaxx/blob/dda2d72654126f33d9cc2b0d66e5f5f1f417172a/web_frontend/osmaxx/excerptexport/templates/excerptexport/templates/list_downloads.html#L18)
     - [ ] [(2)](https://github.com/geometalab/osmaxx/blob/dda2d72654126f33d9cc2b0d66e5f5f1f417172a/web_frontend/osmaxx/excerptexport/templates/excerptexport/partials/download_list.html#L11) and
  - [ ] everywhere else where there's a link to individual downloads.
- A cron job is probably overkill for this. Old files could be identified and deleted when we're about to create new ones

### LINKED ISSUE #553: Remove and cleanup output-files
After a certain duration (days, weeks....) cleanup the output-files and remove them.

Also see #368

hixi:
Reviewed by:
- [x] @das-g 

Closes #368 
Closes #553


Dockerfile.worker
```diff
@@ -113,6 +113,13 @@ RUN mkdir src &&\
 
 ENV CODE /code
 WORKDIR $CODE
+
+# Install dependencies
+RUN wget -O libutf8proc-dev.deb http://ftp.ch.debian.org/debian/pool/main/u/utf8proc/libutf8proc-dev_1.3.1-2_amd64.deb
+RUN wget -O libutf8proc1.deb http://ftp.ch.debian.org/debian/pool/main/u/utf8proc/libutf8proc1_1.3.1-2_amd64.deb
```
das-g: I guess the versions of `libutf8proc-dev` and `libutf8proc` should be kept in sync. Should the version number be extracted to a variable/constant?

osmaxx/excerptexport/_settings.py
```diff
@@ -1,8 +1,17 @@
 import os
+from datetime import timedelta
 
 from django.conf import settings
 
 POLYFILE_LOCATION = os.path.abspath(os.path.join(os.path.dirname(__file__), 'data', 'polyfiles'))
+OLD_RESULT_FILES_REMOVAL_CHECK_INTERVAL_HOURS = timedelta(hours=1)  # default every hour
+PURGE_OLD_RESULT_FILES_AFTER = timedelta(14)  # default to two weeks
```
das-g: Consider passing the days as key word argument: `timedelta(days=14)`

This would remove any doubt or confusion about what time unit is used.

osmaxx/excerptexport/management/commands/purge_old_result_files.py
```diff
@@ -0,0 +1,59 @@
...
+        self._remove_old_files()
+
+    def _success(self, message):
+        self.stdout.write(
+            self.style.SUCCESS(message)
+        )
+
+    def _remove_old_files(self):
+        try:
+            too_old = timezone.now() - PURGE_OLD_RESULT_FILES_AFTER
```
das-g: This constant name reads very nice when setting its value

> Purge old result files after 14 days. &rarr; `PURGE_OLD_RESULT_FILES_AFTER = timedelta(days=14)`

but here, when used in the code, it doesn't play out so well.

Not sure what it should be called instead. `RESULT_FILE_RETAINMENT_DURATION` is the best I can come up with right now.

das-g: If a variable is called `too_old`, I'd expect it to hold a boolean, rather than a datetime.

What you are computing here is the latest possible creation/update time of files that have by now expired. (Or more precisely, as you're comparing with `<`, not with `<=`, the earliest possible creation/update time of files that haven't yet expired.) As neither `latest_possible_update_time_of_now_expired_files` nor `earliest_possible_update_time_of_not_yet_expired_files` make for spiffy variable names and as the expression for computing the variable is rather simple, consider just eliminating it by inlining:
```python
old_files = OutputFile.objects.filter(
    export__updated_at__lt=(timezone.now() - PURGE_OLD_RESULT_FILES_AFTER),
)

File: osmaxx/excerptexport/management/commands/purge_old_result_files.py [hunk 4]
```diff
@@ -0,0 +1,59 @@
...
+            old_files = OutputFile.objects.filter(export__updated_at__lt=too_old)
+            if len(old_files) > 0:
+                for old_file in old_files:
```
das-g: If `len(old_files)` is 0, the `for`-loop will terminate without any iteration, so the `if` here seems completely superfluous.

osmaxx/excerptexport/models/output_file.py
```diff
@@ -21,6 +22,16 @@ class OutputFile(models.Model):
     public_identifier = models.UUIDField(primary_key=False, default=uuid.uuid4, verbose_name=_('public identifier'))
     export = models.OneToOneField(Export, related_name='output_file', verbose_name=_('export'))
 
+    def __str__(self):
+        return \
+            '[' + str(self.id) + '] ' \
+            + ('file: ' + os.path.basename(self.file.name) + ', ' if (self.file and self.file.name) else '') \
+            + 'identifier: ' + str(self.public_identifier)
```
das-g: looks like a job for `str.format` or a template

osmaxx/excerptexport/management/commands/purge_old_result_files.py
hixi: I see what you are getting at - but exactly the name is a bit of an issue, when using this in settings: I set this variable, and then I might want to look for where it is being used in the code. I need first to find that it is being renamed in `_settings.py` to a name that is not even remotely evident from the original settings name.
Even though the code above is not that easy to read in plain English language, I favor this slight annoyance over the renaming.

Also, reading the statements in pseudo-english, I don't find that your proposal is a real improvement to understand:

it is `too_old` when `RESULT_FILE_RETAINMENT_DURATION is being subtracted from now`

compared to what it says now:
it is `too_old` when `PURGE_OLD_RESULT_FILES_AFTER is being subtracted from now`

osmaxx/excerptexport/models/output_file.py
```diff
@@ -36,11 +47,24 @@ def file_extension(self):
...
-        return \
-            '[' + str(self.id) + '] ' \
-            + ('file: ' + os.path.basename(self.file.name) + ', ' if (self.file and self.file.name) else '') \
-            + 'identifier: ' + str(self.public_identifier)
+    def _remove_file(self, save=False):
+        if self.file:
+            file_path = self.file.path
+            file_directory = os.path.dirname(file_path)
+            if os.path.exists(file_directory):
+                shutil.rmtree(file_directory)
```
das-g: Maybe assert that our file is the only file in the directory, so that this won't bite us hard if we ever change the directory structure.

osmaxx/excerptexport/management/commands/purge_old_result_files.py
hixi: I do like the suggestion of in-lining though ;-)

osmaxx/excerptexport/rest_api/serializers.py
```diff
@@ -18,3 +19,9 @@ class Meta:
         model = Excerpt
         fields = ['id', 'excerpt_type', 'color']
         auto_bbox = True
+
+
+class ExportSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Export
+        fields = ['id', 'status']
```
das-g: I like the simplicity here. :+1:

osmaxx/excerptexport/models/output_file.py
hixi: Never use a template in `__str__`, if you get an error, it will not even be able to display the representation of the class it has.

I just moved this line up, where it felt more appropriate, haven't changed anything. This always was a bad string representation for this class, but should not be part of this pull-request. You might want to consider creating an issue for this :wink:

osmaxx/excerptexport/models/output_file.py
hixi: good suggestion :+1:

osmaxx/excerptexport/models/output_file.py
das-g: > I just moved this line up, where it felt more appropriate, haven't changed anything. [Improving this] should not be part of this pull-request.

Oops, I didn't notice that.
:white_check_mark: Leave as is for this PR.

osmaxx/excerptexport/templates/excerptexport/partials/export_display.html
```diff
@@ -12,7 +12,16 @@
...
             </div>
             <div class="col-md-4 col-lg-4">
-                {{ export.status }} at {% firstof export.finished_at|date:"Y-m-d H:i:s" export.updated_at|date:"Y-m-d H:i:s" %}
+                <div class="col-xs-10">
+                    {{ export.status }} at {% firstof export.finished_at|date:"Y-m-d H:i:s" export.updated_at|date:"Y-m-d H:i:s" %}
+                </div>
+                <div class="col-xs-2">
+                    {% if export.status in export.FINAL_STATUSES %}
+                    <div id="delete-export-{{ export.id }}" class="hidden">
+                        <span class="glyphicon glyphicon-remove-circle" aria-hidden="true"></span>
```
das-g: :+1: for using a glyphicon for its _intended_ semantic.

osmaxx/excerptexport/templates/excerptexport/partials/export_display.html
```diff
@@ -24,9 +33,9 @@
                         <a href="{{ export.output_file.file.url }}">{{ export.output_file.get_filename_display }}</a>
                     </p>
                 </div>
-            <div class="col-md-1 col-lg-1">
-                {{ export.output_file.file.size|filesizeformat }}
-            </div>
+                <div class="col-md-1 col-lg-1">
+                    {{ export.output_file.file.size|filesizeformat }}
+                </div>
```
das-g: :+1:

PR #584, opened 2016-06-15
Feature/#582 set status to failed if job is missing on mediator

### LINKED ISSUE #582: Set status to failed if job is not available in mediator
If mediator has no job with the ID in question, set the status of that export to failed. (Should be the case when `ConversionServiceClient` returns 404)

hixi:
Reviewed by:
- [x] @das-g 

Closes #582

### DIFF
--- a/osmaxx/job_progress/middleware.py
+++ b/osmaxx/job_progress/middleware.py
@@ -45,7 +45,7 @@ def update_exports_of_request_user(request):
         except HTTPError as e:
             if hasattr(e.response, 'status_code'):
                 if e.response.status_code == requests.codes['not_found']:
-                    logger.exception("Export #%s doesn't exist on the conversion service anymore.", export.id)
+                    logger.exception("Export #%s doesn't exist on the conversion service.", export.id)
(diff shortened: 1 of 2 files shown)


osmaxx/job_progress/middleware.py
```diff
@@ -39,6 +42,14 @@ def update_exports_of_request_user(request):
     for export in pending_exports:
         try:
             update_export_if_stale(export, request=request)
+        except HTTPError as e:
+            if hasattr(e.response, 'status_code'):
+                if e.response.status_code == requests.codes['not_found']:
+                    logger.exception("Export #%s doesn't exist on the conversion service anymore.", export.id)
```
das-g: "anymore"? It might never have existed, depending on how the inconsistent state occurred. Case in point: Current state of our staging server.

das-g:
otherwise :+1:

osmaxx/job_progress/middleware.py
hixi: you have a point there ;-)

PR #587, opened 2016-06-16
Feature/#581 set status to failed after no update after timeout

### LINKED ISSUE #581: Set export status to failed when no update after timeout
After for example 24h, if no result has been produced, set the export status to failed.

- Use the already existing setting `EXTRACTION_PROCESSING_TIMEOUT_TIMEDELTA`

hixi:
Also cleans up some low hanging fruit.

Reviewed by:
- [x] @das-g 

Closes #581


osmaxx/excerptexport/models/export.py
```diff
@@ -80,11 +80,17 @@ def status_update_url(self):
 
     def set_and_handle_new_status(self, new_status, *, incoming_request):
         assert new_status in dict(self.STATUS_CHOICES)
+        if self.status == new_status and self.last_update_overdue():
+            new_status = self.FAILED
```
das-g: If `set_and_handle_new_status` has been called from `osmaxx.job_progress.views.tracker()`, this doesn't make a lot of sense. So this logic should probably be moved to `osmaxx.job_progress.middleware.update_export()`.

osmaxx/excerptexport/models/export.py
```diff
@@ -80,11 +80,17 @@ def status_update_url(self):
...
+
         if self.status != new_status:
             self.status = new_status
             self.save()
             self._handle_changed_status(incoming_request=incoming_request)
 
+    def last_update_overdue(self):
+        return (self.updated_at + EXTRACTION_PROCESSING_TIMEOUT_TIMEDELTA) < timezone.now()
+
```
das-g: AFAIK, "overdue" doesn't mean "expired" but rather "didn't occur/happen/arrive yet, despite being expected to have occured/happened/arrived already". So it's probably the _next_, not the last update, that's overdue.

das-g: Currently, this we only call this method on exports in non-final statuses. But as the name of the method doesn't reflect that it's only applicable to those, the method should return `False` for exports in a final status, no matter how long ago their last update happened.

osmaxx/excerptexport/models/export.py
hixi: IMHO I think this should always hold true, even if `osmaxx.job_progress.views.tracker()` might be a very unlikely place that a status arrives that is the same as the already existing status.

PR #595, opened 2016-06-20
Feature/ensure removal of old files

hixi:
Not sure where it did go wrong, but this for sure doesn't make it worse...

Reviewed by:
- [x] @das-g


das-g:
:+1:

PR #607, opened 2016-07-04
Feature/#580 calculate and display concrete file size estimations per format

### LINKED ISSUE #580: display actual estimated size for format
Calculate the factor for each format type and display the estimated size on the form for ordering exports from existing excerpts as well as on the form for ordering newly created excerpts.

hixi:
Reviewed by:
- [x] @das-g 

Closes #580


docs/development/project-development-environment.md
```diff
@@ -8,7 +8,7 @@
...
...'dependency_requirements.txt'>various PyPI packages</a> | See [`requirements.txt`](/requirements.txt), [`requirements-local.txt`](/requirements-local.txt) and [`requirements-all.txt`](/requirements-all.txt) | Install in a Python 3 [virtual environment](#dependency_venv). Create one with e.g. <pre class="highlight highlight-source-shell">mkdir -p ~/.virtualenvs && \\<br />virtualenv ~/.virtualenvs/osmaxx -p python3</pre> activate it with <pre class="highlight highlight-source-shell">source ~/.virtualenvs/osmaxx/bin/activate</pre> Then, in the same shell session, use [`pip3`](#dependency_pip3) to install the packages: <pre class="highlight highlight-source-shell"># run from this repo's root dir<br />pip3 install -r requirements-all.txt</pre><hr>`virtualenvwrapper` users can perform all of the above in a single step: <pre class="highlight highlight-source-shell"># run from this repo's root dir<br />mkvirtualenv -r requirements-all.txt \\<br />             -a . -p $(which python3) osmaxx</pre>The `-a .` will also [associate the repo root as the project directory](http://virtualenvwrapper.readthedocs.org/en/latest/command_ref.html#mkvirtualenv) of the new Python 3 virtual environemnt. |
```
das-g: I think, either `requirements-mediator.txt` should be mentioned, too, **or** _only_ `requirements-all.txt` should be mentioned explicitly, e.g.
> See [`requirements-all.txt`](/requirements-all.txt) and the requirements files it refers to.

osmaxx/conversion/converters/converter_garmin/garmin.py
```diff
@@ -18,18 +20,25 @@ def __init__(self, *, out_zip_file_path, area_name, polyfile_string):
         self._osmosis_polygon_file.write(polyfile_string)
         self._osmosis_polygon_file.flush()
         self._polyfile_path = self._osmosis_polygon_file.name
+        self._start_time = None
 
     def create_garmin_export(self):
-        self._to_garmin()
+        self._start_time = timezone.now()
+        unzipped_result_size = self._to_garmin()
```
das-g: While it was not completely baffling (albeit maybe not really intuitive, either) that `._to_garmin()` returned the path to the resulting ZIP file, returning the content's cumulated file size is rather surprising.

As each export to GARMIN format should (and does) use a new `Garmin` instance (otherwise `._start_time` would lose meaning), `_produce_garmin()`, where the size is computed, could store the size on another instance variable instead of returning it.

osmaxx/conversion/converters/utils.py
```diff
@@ -3,6 +3,15 @@
...
 import os
 
+# Use the built-in version of scandir if possible, otherwise
+# use the scandir module version
+try:
+    from os import scandir
+except ImportError:
+    from scandir import scandir
+except:
+    pass
```
das-g: Don't we want to fail here (rather than later at runtime) when both exports were unsuccessful?

osmaxx/conversion/converters/utils.py
```diff
@@ -25,3 +34,11 @@ def zip_folders_relative(folder_list, zip_out_file_path=None):
...
             os.chdir(old_dir)
     return zip_out_file_path
+
+
+def recursive_getsize(path):
+    size = 0
+    for entry in scandir(path):
+        if not entry.name.startswith('.') and entry.is_file():
+            size += os.path.getsize(entry.path)
+    return size
```
das-g: I think this could be written as
```python
return sum(
    os.path.getsize(entry.path)
        for entry in scandir(path)
        if not entry.name.startswith('.') and entry.is_file()
)
```

Not sure which one is easier to grasp.

osmaxx/conversion/management/commands/result_harvester.py
```diff
@@ -41,6 +41,10 @@ def _handle_running_jobs(self):
             self._update_job(job_id=job_id, queue=queue)
 
     def _update_job(self, job_id, queue):
+        if job_id is None:
+            logger.exception("job_id is None, None is not a valid id!")
+            return
```
das-g: `_update_job` is called in loops in `_handle_failed_jobs()` and `_handle_running_jobs()`. Should the loop bodies be wrapped with `try`, so that one `None` in the job IDs doesn't keep all jobs after it in the list from being updated?

das-g: Oops, I kinda assumed an Exception is raised here, but that's not what [`logger.exception`](https://docs.python.org/3.4/library/logging.html#logging.Logger.exception) does. Though, it should only be called from an exception handler (i.e., an `except` clause).

Maybe you meant to just call `logger.error`?

osmaxx/conversion/management/commands/result_harvester.py
```diff
@@ -85,3 +90,16 @@ def add_file_to_job(*, conversion_job, result_zip_file):
         os.makedirs(new_directory_path, exist_ok=True)
     shutil.move(result_zip_file, new_path)
     return new_path
+
+
+def add_meta_data_to_job(*, conversion_job, rq_job):
+    from pbf_file_size_estimation.app_settings import PBF_FILE_SIZE_ESTIMATION_CVS_FILE_PATH
```
das-g: Noticed a typo here, filed PR geometalab/osm-pbf-file-size-estimation-service#3

docs/development/project-development-environment.md
hixi: less text is better, I think.

osmaxx/conversion/converters/utils.py
hixi: I had something similar in the beginning, but switched to this because of readability reasons - but I don't really care which version should get in, of you think the `sum()` is more explicit/readable, I'll include that one.

osmaxx/conversion/management/commands/result_harvester.py
hixi: yeah, `logger.error` make more sense here - it should only show in the logs as exception/error, not actually throw one.

osmaxx/conversion/size_estimator.py
```diff
@@ -0,0 +1,51 @@
...
+        detail_levels.DETAIL_LEVEL_ALL: [115000, 216000, 719000, 1600000],
+        detail_levels.DETAIL_LEVEL_REDUCED: [55000, 66000, 269000, 635000],
+    },
+}
+
+
+def size_estimation_for_format(format_type, detail_level, predicted_pbf_size):
+    import scipy.stats
+    assert format_type in formats.FORMAT_DEFINITIONS.keys()
+    assert detail_level in [level[0] for level in detail_levels.DETAIL_LEVEL_CHOICES]
```
das-g: The call to `keys()` is unnecessary. `format_type in formats.FORMAT_DEFINITIONS` gives the same result and seems to be idiomatic in Python.
For the second assertion, a set (instead of a list) would match the intended semantics better, probably.

As `size_estimation_for_format()` does nothing with `format_type` and `detail_level` except passing them on to `get_data()`, I wonder whether these assertions should be moved there.

osmaxx/conversion/size_estimator.py
```diff
@@ -0,0 +1,51 @@
...
+    x, y = get_data(format_type, detail_level)
```
das-g: Maybe use content-describing variable names here. The named arguments on the next line are enough to see what role they take in the linear regression.

osmaxx/conversion/converters/utils.py
hixi: :+1:

osmaxx/excerptexport/static/excerptexport/scripts/estimate_size.js
```diff
@@ -0,0 +1,62 @@
...
+            '/api/estimated_file_size/',
+            {
+                'north': north,
+                'east': east,
+                'south': south,
+                'west': west
+            });
+    };
+
+    window.estimateSizeForFormats = function (pbfSize, detailLevel) {
```
das-g: Not used outside this module. Can be made "private" if wanted.

requirements-all.txt
```diff
@@ -0,0 +1,4 @@
+# just a meta file to contract all requirements in one file
+-r requirements.txt
+-r requirements-mediator.txt
+-r requirements-local.txt
```
das-g: :+1:

PR #619, opened 2016-07-07
return a 503 (Service unavailable) for unknown domain name

hixi:
return a 503 (Service unavailable) when an unknown domain name is passed along.

Reviewed by:
- [x] @das-g

### DIFF
--- a/Dockerfile.nginx
+++ b/Dockerfile.nginx
@@ -1,2 +1,4 @@
 FROM nginx:alpine
-COPY ./docker_entrypoint/nginx/default.conf /etc/nginx/conf.d/default.conf
+COPY ./docker_entrypoint/nginx/default.conf.template /etc/nginx/conf.d/default.conf.template
+CMD DOMAIN_NAMES=$(echo $VIRTUAL_HOST | sed 's/,/ /g') envsubst '$DOMAIN_NAMES' < /etc/nginx/conf.d/default.conf.template \
+    && DOMAIN_NAMES=$(echo $VIRTUAL_HOST | sed 's/,/ /g') envsubst '$DOMAIN_NAMES' < /etc/nginx/conf.d/default.conf.template > /etc/nginx/conf.d/default.conf && nginx -g 'daemon off;'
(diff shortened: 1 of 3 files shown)


Dockerfile.nginx
```diff
@@ -1,2 +1,4 @@
 FROM nginx:alpine
-COPY ./docker_entrypoint/nginx/default.conf /etc/nginx/conf.d/default.conf
+COPY ./docker_entrypoint/nginx/default.conf.template /etc/nginx/conf.d/default.conf.template
+CMD DOMAIN_NAMES=$(echo $VIRTUAL_HOST | sed 's/,/ /g') envsubst '$DOMAIN_NAMES' < /etc/nginx/conf.d/default.conf.template \
+    && DOMAIN_NAMES=$(echo $VIRTUAL_HOST | sed 's/,/ /g') envsubst '$DOMAIN_NAMES' < /etc/nginx/conf.d/default.conf.template > /etc/nginx/conf.d/default.conf && nginx -g 'daemon off;'
```
das-g: consider `cat`ing the resulting file instead of doing the substitution twice (once for output, once for file)

docker_entrypoint/nginx/default.conf.template
```diff
@@ -39,13 +39,20 @@ proxy_set_header Upgrade $http_upgrade;
 proxy_set_header X-Real-IP $remote_addr;
 proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
 
+server {
+    server_name _; # This is just an invalid value which will never trigger on a real hostname.
+    listen 80;
+    access_log /var/log/nginx/access.log;
+    return 503;
```
das-g: Would [`421` 'Misdirected Request'](https://tools.ietf.org/html/rfc7540#section-9.1.2) make more sense?

hixi: 421 can be cached, therefore I'd say `nay`.

Dockerfile.nginx
hixi: good idea :-)

PR #620, opened 2016-07-07
remove OutputFile DB object even without attached file

### LINKED ISSUE #613: Error on File Cleanup: The 'file' attribute has no file associated with it.
https://sentry.njo.ch/geometalab/osmaxxprod/issues/175/

```
ValueError: The 'file' attribute has no file associated with it.
  File "osmaxx/excerptexport/management/commands/purge_old_result_files.py", line 57, in _remove_old_files
    path = old_file.file.path
```

hixi:
check for an available file on the output file, otherwise it might raise an error, since `__file__isnull=True` does not cover that case

Reviewed by:
- [x] @das-g 

Closes #613


tests/excerptexport/management/test_purge_old_result_files.py
```diff
@@ -24,6 +24,21 @@ def output_file_with_file_too_old(db, output_file_with_file, mocker):
     return output_file_with_file
 
 
+@pytest.fixture
+def output_file_too_old(db, output_file, mocker):
```
das-g: `mocker` fixture injected but not used

das-g: "too old" for what? Also, this fixture provides a file (model) object, not a boolean, so the name should _end_ in "file". Maybe `expired_file`?

hixi: good idea, same applies above (unchanged code).

tests/excerptexport/management/test_purge_old_result_files.py
```diff
@@ -24,6 +24,21 @@ def output_file_with_file_too_old(db, output_file_with_file, mocker):
...
+    from osmaxx.excerptexport._settings import RESULT_FILE_AVAILABILITY_DURATION
+    just_a_bit_too_old = timezone.now() - RESULT_FILE_AVAILABILITY_DURATION - timezone.timedelta(minutes=1)
+
+    output_file.file_removal_at = just_a_bit_too_old
+    output_file.save()
+
+    output_file.refresh_from_db()
+
+    assert output_file.file_removal_at == just_a_bit_too_old
```
das-g: Is this assertion there to verify you've called `save()`? Otherwise, checking for the value you just assigned doesn't seem to serve any purpose.

tests/excerptexport/management/test_purge_old_result_files.py
```diff
@@ -24,6 +24,21 @@ def output_file_with_file_too_old(db, output_file_with_file, mocker):
     return output_file_with_file
 
 
+@pytest.fixture
+def output_file_too_old(db, output_file, mocker):
+    from osmaxx.excerptexport._settings import RESULT_FILE_AVAILABILITY_DURATION
+    just_a_bit_too_old = timezone.now() - RESULT_FILE_AVAILABILITY_DURATION - timezone.timedelta(minutes=1)
+
+    output_file.file_removal_at = just_a_bit_too_old
```
das-g:
```python
output_file.file_removal_at = timezone.now() - timezone.timedelta(minutes=1)
```
should suffice for the file being expired. Why also subtract `RESULT_FILE_AVAILABILITY_DURATION`?

tests/excerptexport/management/test_purge_old_result_files.py
hixi: Using `refresh_from_db()` I want to be sure that my `file_removal_at` is correct in the database, if we change the code (I did while working on it), the error is very hard to discover.

tests/excerptexport/management/test_purge_old_result_files.py
hixi: :+1:

osmaxx/excerptexport/management/commands/purge_old_result_files.py
```diff
@@ -54,10 +54,14 @@ def _remove_old_files(self):
                 export__output_file__file__isnull=False,
             )
             for old_file in old_files:
-                path = old_file.file.path
                 pk = old_file.id
+                success_message = "Output File #{} removed".format(pk)  # default
+                # if a file has never been attached, but still is stale, remove the object.
```
das-g: This comment seems misplaced. If you want, replace it with a comment directly above `old_file.delete()` saying something like
```python
# Remove object, no matter whether a on-disk file has been attached or not.
```

(Though, the comment could just as well be removed without replacement, I think.)

tests/excerptexport/management/test_purge_old_result_files.py
das-g: :white_check_mark: makes sense, then

osmaxx/excerptexport/management/commands/purge_old_result_files.py
hixi: based on the times we removed this exact line of code because it seemed unneeded, I made this comment. But now that we have a test for this scenario, removing makes sense :-)

osmaxx/excerptexport/management/commands/purge_old_result_files.py
```diff
@@ -54,10 +54,13 @@ def _remove_old_files(self):
                 export__output_file__file__isnull=False,
```
das-g: should we use "expired" instead of "old" in the names here, too?

hixi: by names you mean: source filename, variable name and function name, right?

das-g: Didn't notice the source filename, but now that you mention it, yes, that too.
Change all the names

Issue #578, opened 2016-06-15
Show only suitable UTM Zones


hixi:
For the selected area, create a service, so that:

- [ ] sending an bounding box returns only the available UTM-Zones for that area
- [ ] using Javascript, the selectable list is being reduced to those choices

PR #622, opened 2016-07-07
Where should the code added on feature/#578_UTM_zone_tools live?

das-g:
**Do not merge**

For now, for discussing what to do with this. (Might be turned into a merge request later, depending on result.)

This new code for #578 is fairly independent of the rest of the project. Should it live within the `osmaxx` repo or in a separate (and POPyO) PyPI package?

Some of the tests would have to remain in `osmaxx` as they depend on PostGIS, lest we want the PyPI package to have that as a test dependency, too.

**Do not merge**


hixi:
When you merge develop into your branch, the slow tests should work as well.

hixi:
IMHO think this is a good candidate for an external package.

I'd make an PyPI-Package (using [twine](https://docs.python.org/3/distributing/index.html#installing-the-tools) to [push to PyPI](https://packaging.python.org/distributing/#uploading-your-project-to-pypi)).

I see two options (and a combination thereof):
1. making it an independent package, no connection to django/django-rest-framework
2. making it a [django-rest-framework-package](http://www.django-rest-framework.org/topics/third-party-resources/)
3. doing both, and using the first as a dependency on the second

I would favor option 2: Having it ready as a REST-Service, this can easily be included in a Dockerimage for simple deployment and be used for the OSMaxx project right away, but doesn't limit usage in any other way (one should easily be able to use the package without installing rest-framework, if the viewsets aren't being used).

conftest.py
```diff
@@ -1,3 +1,5 @@
 def pytest_addoption(parser):
     parser.addoption("--runslow", action="store_true",
                      help="run slow tests")
+    parser.addoption("--all-utm-zones", action="store_true",
```
hixi: instead of running the test with really all utm zones, I wonder if it would make sense to test the edge cases, boundary cases and happy path cases only. This would require more thought on what those cases actually are, but just because we `can` test all cases, doesn't mean we necessarily `need` to (they seem to me quite similar in what they test).

osmaxx/geodesy/coordinate_reference_system.py
```diff
@@ -0,0 +1,99 @@
...
+from django.contrib.gis.geos.polygon import Polygon
+
+from osmaxx.conversion_api.coordinate_reference_systems import WGS_84
+
+MIN_LONGITUDE_DEGREES = -180
+MAX_LONGITUDE_DEGREES = +180
+
+
+class UniversalTransverseMercatorZone:
+    HEMISPHERE_PREFIXES = dict(
```
hixi: I think in this case the extra lines do not enhance readability: `dict(north=326, south=327)` is easy enough to read and understand.

osmaxx/geodesy/coordinate_reference_system.py
```diff
@@ -0,0 +1,99 @@
...
+        north=326,
+        south=327,
+    )
+    NUMBER_OF_ZONES_PER_HEMISPHERE = 60
+    VALID_ZONE_NUMBERS = range(1, NUMBER_OF_ZONES_PER_HEMISPHERE + 1)
+    ZONE_WIDTH_DEGREES = 360 / NUMBER_OF_ZONES_PER_HEMISPHERE
+    MAX_LONGITUDE_OFFSET = 90.0 - 9.9e-14
```
hixi: what is `9.9e-14`?

osmaxx/geodesy/coordinate_reference_system.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def __init__(self, hemisphere, utm_zone_number):
+        assert hemisphere in self.HEMISPHERE_PREFIXES.keys()
```
hixi: No need to use `.keys()` here.

osmaxx/geodesy/coordinate_reference_system.py
```diff
@@ -0,0 +1,99 @@
...
+            # cut at idealized international date line
+            return MultiPolygon(
+                Polygon.from_bbox((xmin, ymin, MAX_LONGITUDE_DEGREES, ymax)),
+                Polygon.from_bbox((MIN_LONGITUDE_DEGREES, ymin, xmax, ymax)),
+                srid=WGS_84,
+            )
+
+    @property
+    def srid(self):
+        return self.HEMISPHERE_PREFIXES[self.hemisphere] * 100 + self.utm_zone_number
```
hixi: interesting way of computing the srid. :smile:

osmaxx/geodesy/coordinate_reference_system.py
```diff
@@ -0,0 +1,99 @@
...
+
+    @property
+    def central_meridian_longitude_degrees(self):
+        return MIN_LONGITUDE_DEGREES + (self.utm_zone_number - 0.5) * self.ZONE_WIDTH_DEGREES
```
hixi: What does the subtraction of `0.5` do (or why is it needed)?

osmaxx/geodesy/coordinate_reference_system.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def __eq__(self, other):
```
hixi: :+1:

osmaxx/geodesy/coordinate_reference_system.py
das-g: We have two GEOS/Proj.4 implementations/versions involved: The one GeoDjango uses and the one PostGIS uses.

In my tests, PostGIS was able to transform WGS-84 coordinates to UTM iff the longitude was within up to 90° of the chosen UTM zone's "central meridian". However, with GeoDjango, it worked only _nearly_ up to 90°, at least on my machine. The difference from 90° (9.9&times;10<sup>-14</sup>°) was found by trial and error. Are more precise value could probably be found by systematic (e.g. automated) bisection, but I deemed the value I found close enough for our purposes.

osmaxx/geodesy/coordinate_reference_system.py
```diff
@@ -0,0 +1,99 @@
...
+    def __hash__(self):
+        return hash((self.hemisphere, self.utm_zone_number))
+
+    def __str__(self):
+        return "UTM Zone {zone_number}, {hemisphere}ern hemisphere".format(
+            zone_number=self.utm_zone_number,
+            hemisphere=self.hemisphere,
+        )
+
+    def __repr__(self):
```
hixi: From the [__repr__ documentation](https://docs.python.org/3/reference/datamodel.html#object.__repr__): I think this should return `UniversalTransverseMercatorZone({hemisphere}, {zone_number})`.

I know, there is a shorthand below, but this class shouldn't know about that.

osmaxx/geodesy/coordinate_reference_system.py
```diff
@@ -0,0 +1,99 @@
...
+        return "UTMZone({hemisphere}, {zone_number})".format(
+            zone_number=self.utm_zone_number,
+            hemisphere=repr(self.hemisphere),
+        )
+
+UTMZone = UniversalTransverseMercatorZone
+UTM_ZONE_NUMBERS = UTMZone.VALID_ZONE_NUMBERS
+ALL_UTM_ZONES = frozenset(UTMZone(hs, nr) for hs in UTMZone.HEMISPHERE_PREFIXES for nr in UTM_ZONE_NUMBERS)
```
hixi: I think this would benefit from newlines for readability:

```python
frozenset(
    UTMZone(hs, nr) 
    for hs in UTMZone.HEMISPHERE_PREFIXES 
    for nr in UTM_ZONE_NUMBERS
)
```

osmaxx/geodesy/coordinate_reference_system.py
```diff
@@ -0,0 +1,99 @@
...
+
+def utm_zones_for_representing(geom):
+    return frozenset(zone for zone in ALL_UTM_ZONES if zone.can_represent(geom))
+
+
+def wrap_longitude_degrees(longitude_degrees):
+    return confine(longitude_degrees, MIN_LONGITUDE_DEGREES, MAX_LONGITUDE_DEGREES)
+
+
+def confine(value, lower_bound, upper_bound):
```
hixi: Good name IMO.

osmaxx/geodesy/coordinate_reference_system.py
das-g: Would string concatenation (and conversion back to `int`) make more sense? (What I'm doing here is probably better performant, but that shouldn't matter as this isn't called in high frequency.)

tests/inside_worker_test/air_traffic_test.py
```diff
@@ -4,7 +4,7 @@
 import sqlalchemy
 from sqlalchemy.sql.schema import Table as DbTable
 
-from tests.inside_worker_test.conftest import slow
+from tests.utils import slow
```
hixi: :+1: for throwing this into the test utils package.

osmaxx/geodesy/coordinate_reference_system.py
hixi: This isn't something I can understand by just seeing the value.
I think this would be one place where a comment wouldn't hurt (since putting the entire explanation into a variable seems to generate a bit a longish name :laughing: ).

osmaxx/geodesy/coordinate_reference_system.py
hixi: No, I don't think so. At least I understand without any further explanation what this does.

osmaxx/geodesy/coordinate_reference_system.py
das-g: UTM Zone 1 goes from longitude (-180°) to longitude (-180° + zone width). Its central meridian is centered between these boundaries.

- 180° + (1 - 0) &times; zone width gives us the east edge of zone 1.
- 180° + (1 - &half;) &times; zone width gives us the central meridian of zone 1.
- 180° + (1 - 1) &times; zone width gives us the west edge of zone 1.

where the first `1` is the zone number.

Not all other zones have the same width (at least [not everywhere](http://earth-info.nga.mil/GandG/coordsys/grids/universal_grid_system.html#zz4)), but their "central meridians" are defined consistent with that formula (and thus not always actually central).

osmaxx/geodesy/coordinate_reference_system.py
das-g: How does using the alias class name violate
> [T]his should look like a valid Python expression that could be used to recreate an object with the same value (given an appropriate environment).

in any way? Or is it in opposition to something else in the documentation?

osmaxx/geodesy/coordinate_reference_system.py
hixi: Would this warrant a comment?

osmaxx/geodesy/coordinate_reference_system.py
das-g: Kinda unspecific for what it does, IMHO.

Let's see if http://math.stackexchange.com/questions/1851903/does-this-modulo-based-function-have-a-name brings up a better one.

osmaxx/geodesy/coordinate_reference_system.py
hixi:
```python
UniversalTransverseMercatorZone('north', 32).__repr__
> UTM('north', 32)
```
Doesn't seem very logical to me to use the alias as `__repr__`, when the class definition doesn't know it's being aliased.

osmaxx/geodesy/coordinate_reference_system.py
das-g: Only if we can think of a good one.

Alternatively, the geometric interpretation of the formula is maybe more apparent if written as
```python
return MIN_LONGITUDE_DEGREES \
    + (self.utm_zone_number - 1) * self.ZONE_WIDTH_DEGREES \
    + self.ZONE_WIDTH_DEGREES / 2
```

hixi: or a docstring:
```
"""
blah blah blah...

UTM Zone 1 goes from longitude (-180°) to longitude (-180° + zone width). Its central meridian is centered between these boundaries.

180° + (1 - 0) × zone width gives us the east edge of zone 1.
180° + (1 - ½) × zone width gives us the central meridian of zone 1.
180° + (1 - 1) × zone width gives us the west edge of zone 1.
where the first 1 is the zone number.
"""
```

osmaxx/geodesy/coordinate_reference_system.py
```diff
@@ -7,18 +7,21 @@
...
 
 
-class UniversalTransverseMercatorZone:
-    HEMISPHERE_PREFIXES = dict(
-        north=326,
-        south=327,
-    )
+class UTMZone:
+    """
+    Universal Transverse Mercator (UTM) zone
```
hixi: :+1:

osmaxx/geodesy/coordinate_reference_system.py
```diff
@@ -0,0 +1,106 @@
...
+
+from osmaxx.conversion_api.coordinate_reference_systems import WGS_84
+
+MIN_LONGITUDE_DEGREES = -180
+MAX_LONGITUDE_DEGREES = +180
+
+
+class UTMZone:
+    """
+    Universal Transverse Mercator (UTM) zone
```
hixi: :+1:

das-g:
DRF package it is: https://github.com/geometalab/drf-utm-zone-info

PR #626, opened 2016-07-11
use different version to test on jenkins; remove django 1.8 support

hixi:
Reviewed by:
- [x] @das-g


tox.ini
```diff
@@ -2,16 +2,19 @@
 envlist =
        ; disable creating docs until https://github.com/mkdocs/mkdocs/issues/807 is resolved
        ; py34-{flake8,docs}
```
das-g: mkdocs/mkdocs#807 has been closed. Can we re-enable the `docs` build?

hixi: before we enable them, we should test that locally - I'm not even sure they still work.

PR #642, opened 2016-07-13
merge work after merge with develop

hixi:
Reviewed by:
- [x] @das-g


osmaxx/conversion/migrations/0010_auto_20160713_1036.py
```diff
@@ -0,0 +1,25 @@
...
+
+    dependencies = [
+        ('conversion', '0009_merge'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='job',
+            name='queue_name',
+            field=models.CharField(choices=[('default', 'default'), ('high', 'high')], default='default', help_text='queue name for processing', max_length=50, verbose_name='queue name'),
```
das-g: The order of choices is non-reproducible, because [it is produced by iterating over](https://github.com/geometalab/osmaxx/blob/5751ad5936e4ef38c25bd91d6c71be14fabfdf1f/osmaxx/conversion/models.py#L57) `dict` [`settings.RQ_QUEUES`](https://github.com/geometalab/osmaxx/blob/5751ad5936e4ef38c25bd91d6c71be14fabfdf1f/osmaxx_conversion_service/config/settings/common.py#L254).

Either [`sort`](https://docs.python.org/library/functions.html#sorted) for the iteration or use an [`OrderedDict`](https://docs.python.org/library/collections.html#collections.OrderedDict).

hixi: :+1:

PR #644, opened 2016-07-13
also add the exclusive worker to development yml

hixi:
Reviewed by:
- [x] @das-g

### DIFF
--- a/docker-compose-dev.yml
+++ b/docker-compose-dev.yml
@@ -70,6 +70,25 @@ services:
       driver: "json-file"
       options:
         max-size: 1M
+  worker-exclusive:
+    extends:
+      file: docker-compose-common.yml
+      service: conversionbase
+    volumes:
+      - ./osmaxx:/home/py/osmaxx
+      - ./web_frontend:/home/py/web_frontend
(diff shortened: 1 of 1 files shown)


docker-compose-dev.yml
```diff
@@ -70,6 +70,25 @@ services:
...
+      - ./osmaxx:/home/py/osmaxx
+      - ./web_frontend:/home/py/web_frontend
+      - ./osmaxx_conversion_service:/home/py/osmaxx_conversion_service
+    build:
+      context: .
+      dockerfile: Dockerfile.worker
+    command: [honcho, -f, ./osmaxx_conversion_service/Procfile.worker, start]
+    environment:
+      - DJANGO_SECRET_KEY=insecure!4
+      - DJANGO_SETTINGS_MODULE=osmaxx_conversion_service.config.settings.local
```
das-g: Shouldn't this one listen to a different set of queues than `worker`?

das-g:
:+1:

docker-compose-dev.yml
hixi: no, since this file is overwriting and adding the docker-compose.yml file when used as specified (`docker-compose -f docker-compose.yml -f docker-compose-dev.yml`)

PR #646, opened 2016-07-13
capitalize also emails

### LINKED ISSUE #629: Capitalization of proper names
@sfkeller noted:

* We still have occurences of "Osmaxx" instead of "OSMaxx" in user-visible strings (e.g. in the `<title>` HTML element)
* Esri, although being an acronym, has switched to capitalize their name "Esri" instead of the previous "ESRI".
* "Garmin" should be used instead of "GARMIN" or "garmin" (the latter is currently used on the GIS options title)

Use the current official capitalization everywhere. (For user-visible strings.)

hixi:
Reviewed by:
- [x] @das-g 

Closes #629


docs/development/project-development-environment.md
```diff
@@ -216,7 +216,7 @@ The following list ist supported:
     - DJANGO_EMAIL_HOST_PASSWORD=''
     - DJANGO_EMAIL_HOST_USER=''
     - DJANGO_EMAIL_PORT=25
-    - DJANGO_EMAIL_SUBJECT_PREFIX='[OSMAXX-DEV] '
+    - DJANGO_EMAIL_SUBJECT_PREFIX='[OSMaxx-dev] '
```
das-g: Good find! :+1:

osmaxx/utilities/shortcuts.py
```diff
@@ -32,7 +32,7 @@ def inform_mail(self, subject, mail_body, warn_if_no_email=True):
         try:
             email_address = self.recipient.email
             mail.send_mail(
-                '[OSMAXX] ' + subject,
+                '[OSMaxx] ' + subject,
```
das-g: Should `settings.EMAIL_SUBJECT_PREFIX` be used here?

hixi: either that, or remove the setting.

das-g: I'd lean towards using the setting here. Then, if we ever use [`mail_admins()`](https://docs.djangoproject.com/en/1.9/topics/email/#mail-admins) or [`mail_managers()`](https://docs.djangoproject.com/en/1.9/topics/email/#mail-managers), (where Django would make use of the setting automatically), the setting is already in place.

hixi: Then let's do that.

PR #647, opened 2016-07-14
Feature/push master when releasing

das-g:
* push `master` branch, too
* push only newly created tag for the new release, not all local tags

This assumes that the target repository is `origin`. (Previous assumption was that the target repo for `develop` was the repo tracked by `develop` and that the target repository for the tags was `origin`.)

### Reviewed by
- [x] @hixi


hixi:
:+1:

PR #648, opened 2016-07-14
introduce casting function

### LINKED ISSUE #645: sqlalchemy.exc.DataError: (psycopg2.DataError) invalid input syntax for type double precision: "a"
https://sentry.njo.ch/geometalab/osmaxxstaging/issues/197/

```
DataError: (psycopg2.DataError) invalid input syntax for type double precision: "a"
 [SQL: 'INSERT INTO osmaxx.building_a\n  SELECT osm_id as osm_id,\n    osm_timestamp as lastchange ,\n    CASE\n     WHEN osm_id<0 THEN \'R\' -- R=Relation\n     ELSE \'W\'               -- W=Way\n     END AS geomtype,\n    ST_Multi(way) AS geom,\n    building as type,\n    name as name,\n    "name:en" as name_en,\n    "name:fr" as name_fr,\n    "name:es" as name_es,\n    "name:de" as name_de,\n    int_name as name_int,\n    case\n        when name is not null AND name = osml10n_translit(name) then name\n        when "name:en" is not null then "name:en"\n        when "name:fr" is not null then "name:fr"\n        when "name:es" is not null then "name:es"\n        when "name:de" is not null then "name:de"\n        when name is not null then osml10n_translit(name)\n        else NULL\n    end as label,\n    cast(tags as text) as tags,\n    cast(nullif(height,\'\') as float) as height\n  FROM osm_polygon\n  WHERE building is not null;\n']
(16 additional frame(s) were not displayed)
...
  File "osmaxx/conversion/converters/converter_gis/bootstrap/bootstrap.py", line 30, in bootstrap
    self._filter_data()
  File "osmaxx/conversion/converters/converter_gis/bootstrap/bootstrap.py", line 79, in _filter_data
    self._execute_sql_scripts_in_folder(script_folder_path)
  File "osmaxx/conversion/converters/converter_gis/bootstrap/bootstrap.py", line 106, in _execute_sql_scripts_in_folder
    self._postgres.execute_sql_file(script_path)
  File "osmaxx/conversion/converters/converter_gis/helper/postgres_wrapper.py", line 27, in execute_sql_file
    return self.execute_sql_command(psql_command_file.read())
  File "osmaxx/conversion/converters/converter_gis/helper/postgres_wrapper.py", line 35, in execute_sql_command
    result = connection.execute(sqlalchemy.text(sql))
```

hixi:
Reviewed by:
- [x] @das-g 

Re-reviewed by (starting at commit 5f9da62):
- [x] @das-g 

Closes #645


osmaxx/conversion/converters/converter_gis/bootstrap/sql/functions/0060_cast_to_float_or_null.sql
```diff
@@ -0,0 +1,10 @@
+-- try casting to float and return NULL if failed
+create or replace function cast_to_float_null_if_failed(v_input text) returns float as $$
+begin
+    return cast(v_input as float);
+exception
+    when invalid_text_representation then
+        RAISE NOTICE 'Invalid float value: "%".  Returning NULL.', v_input;
+        return NULL;
+end;
+$$ language plpgsql immutable;
```
das-g: I know our current scripts have varying mixtures, but please be consistent at least for new scripts: Please either write all SQL and PL/SQL keywords in all-caps or write all of them all-lowercase.

tests/inside_worker_test/cast_to_float_or_null_test.py
```diff
@@ -0,0 +1,42 @@
...
+
+from tests.inside_worker_test.conftest import slow
+
+
+@pytest.fixture(params=[2, 2.2, 3.898986, "3.898986", "6", "0.2", 0.6])
+def valid_float_representation(request):
+    return request.param
+
+
+@pytest.fixture(params=["a2", "10b", "3.898986k", "3k.898986", "l6.9"])
```
das-g: The last param is easily misread for `16.9`, so that makes this fixture a bit confusing.

Might also be good to have examples with whitespace or with decimal separators of other locales.

tests/inside_worker_test/cast_to_float_or_null_test.py
```diff
@@ -0,0 +1,42 @@
+import pytest
+import sqlalchemy
+
+from tests.inside_worker_test.conftest import slow
+
+
+@pytest.fixture(params=[2, 2.2, 3.898986, "3.898986", "6", "0.2", 0.6])
```
das-g: Examples with negative numbers and with scientific notation wouldn't hurt. Not sure why to have both floats and strings here, as you convert them to strings for the test anyway (and to floats for the assertion).

tests/inside_worker_test/cast_to_float_or_null_test.py
```diff
@@ -0,0 +1,42 @@
...
+
+@slow
+def test_cast_to_float_null_if_failed_returns_floats_with_valid_floats(osmaxx_functions, valid_float_representation):
+    engine = osmaxx_functions
+    result = engine.execute(
+        sqlalchemy.text(
+            "select cast_to_float_null_if_failed($${}$$) as float_value;".format(valid_float_representation)
+        ).execution_options(autocommit=True)
+    )
+    assert result.rowcount == 1
```
das-g: If this assertion fails, `fetchall` below won't be executed and `result` thus not closed.

tests/inside_worker_test/cast_to_float_or_null_test.py
hixi: supporting `,` additionally to `.` I think is a good idea, but do we really want to support the thousand separators and similar awkwardness?

das-g: I meant adding them as negative examples (params of `invalid_floats`).

hixi: ah :+1:

tests/inside_worker_test/cast_to_float_or_null_test.py
hixi: We actually have that at [other places](https://github.com/geometalab/osmaxx/blob/10783e7bea24c384e08ac39cb7f122de6d47a7e0/tests/inside_worker_test/transliterate_sql_function_learning_test.py#L28) as well, but you're right, we should've taken care of this.

tests/inside_worker_test/conftest.py
```diff
@@ -85,6 +85,7 @@ def osmaxx_functions(osm_tables):
         'sql/functions/0030_transliterate.sql',
         'sql/functions/0040_interpolate_addresses.sql',
         'sql/functions/0050_cast_to_int.sql',
+        'sql/functions/0060_cast_to_float_or_null.sql',
```
hixi: @das-g is there a reason we keep [two](https://github.com/geometalab/osmaxx/pull/648/files#diff-6b9af5ce056da868f1e31805af76335cR201) lists of these sql scripts in the tests?

das-g: Historic reasons: Well, I guess yeah.
Good reasons: Not aware of any.

tests/inside_worker_test/label_sql_test.py
```diff
@@ -1193,8 +1195,8 @@ def test_label_water_l(osmaxx_schemas, label_input):
...
     address_script = 'sql/filter/water/030_water_l_insert_table.sql'
-    result = engine.execute(sqlalchemy.text(sql_from_bootstrap_relative_location(address_script)).execution_options(autocommit=True))
-    assert result.rowcount == 1
+    with closing(engine.execute(sqlalchemy.text(sql_from_bootstrap_relative_location(address_script)).execution_options(autocommit=True))) as result:
+        assert result.rowcount == 1
 
-    result = engine.execute(sqlalchemy.text("select label from osmaxx.water_l").execution_options(autocommit=True))
-    assert result.fetchone()['label'] == expected_label
+    with closing(engine.execute(sqlalchemy.text("select label from osmaxx.water_l").execution_options(autocommit=True))) as result:
+        assert result.fetchone()['label'] == expected_label
```
das-g: :+1:

PR #652, opened 2016-07-18
Feature/fix recursive size calculation

hixi:
Reviewed by:
- [x] @das-g


osmaxx/conversion/migrations/0011_remove_falsely_zeroed_unzipped_result_size_values_20160718_1445.py
```diff
@@ -0,0 +1,30 @@
...
+from django.db import migrations
+from django.db.migrations import RunPython
+
+
+def remove_unzipped_result_size_that_are_zero(apps, schema_editor):
+    Job = apps.get_model("conversion", "Job")  # noqa
+    for job in Job.objects.filter(unzipped_result_size=0.0):
+        job.unzipped_result_size = None
+        job.estimated_pbf_size = None
+        job.save()
```
das-g: :+1:

tests/conversion/utils_test.py
```diff
@@ -0,0 +1,37 @@
...
+    request.addfinalizer(rm_tmpdir)
+
+    file_sizes = [102, 304, 10, 2000, 10, 678691, 1667, 1254]
+    expected_size = sum(file_sizes)
+    subdir = tempdir.name
+    for index, file_size in enumerate(file_sizes):
+        if index % 2 == 0:
+            subdir = os.path.join(subdir, str(file_size))
+            os.mkdir(subdir)
+        elif index != 0:
```
das-g: `else` would do the same (here) as `elif index != 0`:  If `index % 2 != 0`, then `index != 0` is guaranteed. (But maybe you intended something else?)

tests/conversion/utils_test.py
```diff
@@ -0,0 +1,37 @@
...
+    def rm_tmpdir():
+        shutil.rmtree(tempdir.name)
+    request.addfinalizer(rm_tmpdir)
+
+    file_sizes = [102, 304, 10, 2000, 10, 678691, 1667, 1254]
+    expected_size = sum(file_sizes)
+    subdir = tempdir.name
+    for index, file_size in enumerate(file_sizes):
+        if index % 2 == 0:
+            subdir = os.path.join(subdir, str(file_size))
```
das-g: A hardcoded list of file paths (or a dictionary mapping these paths to file sizes, or a list of 2-tuples) would be much easier to comprehend than iteratively modifying `subdir` in this `for`-loop.

tests/conversion/utils_test.py
```diff
@@ -0,0 +1,37 @@
...
+            subdir = os.path.join(subdir, str(index))
+            os.mkdir(subdir)
+        file_path = os.path.join(subdir, 'example_{}.file'.format(str(file_size)))
+        with open(file_path, 'w') as file:
+            file.truncate(file_size)
+        assert os.path.getsize(file_path) == file_size
+    return tempdir.name, expected_size
+
+
+def test_recursive_getsize_does_count_subdirectories(directory_with_subdirectories):
```
das-g: "count" is the wrong verb here (as the result is sum of file sizes, not the number of subdirectories).

tests/conversion/utils_test.py
hixi: True, you think the explicitness is overstated here?

tests/conversion/utils_test.py
hixi: `sums_up_sizes_correctly`?

tests/conversion/utils_test.py
hixi: That's just a lot of code doing not helping to reduce repetitive work. What is the problem that you find hard to comprehend?

tests/conversion/utils_test.py
das-g: I can't think of anything better right now, so go with that.

tests/conversion/utils_test.py
das-g: It's not "hard" _per se_. I just had to read this loop very carefully to figure out what it's doing (all the while guessing why it's doing what it's doing), whereas if the loop was
```python
for dir_path, filename, file_size in files_to_create_with_sizes:
    os.mkdir(path, exists_ok=True)
    file_path = os.path.join(dir_path, filename)
    with open(file_path, 'w') as file:
        file.truncate(file_size)
    assert os.path.getsize(file_path) == file_size
```
just skimming over it would have sufficed to see what's going on.

Populating `files_to_create_with_sizes` explicitly (with literals) would be rather verbose, but would reveal the resulting structure almost visually:

```python
files_to_create_with_sizes = [
    ('' 'a.txt', 102),
    ('b' 'c.txt', 304),
    ('b' 'd.txt', 10),
    ('b/e' 'f.txt', 2000),
    ('g/h/i/j' 'k.txt', 10),
]

PR #670, opened 2016-08-17
Feature/#663 validate email adresses

### LINKED ISSUE #663: Validate email addresses
When a (new or existing) user changes his E-Mail address, validate it.

hixi:
Reviewed by:
- [x] @das-g 

Closes #663


osmaxx/profile/forms.py
```diff
@@ -1,12 +1,14 @@
...
 from django.core.urlresolvers import reverse
 from django import forms
+from django.utils.translation import ugettext_lazy as _
+
+from osmaxx.profile.models import Profile
 
 
 class ProfileForm(forms.ModelForm):
-    email = forms.EmailField(max_length=200, required=True)
+    unverified_email = forms.EmailField(max_length=200, required=True, label=_('email'))
```
das-g: Shouldn't the label be "Email **address**"?

osmaxx/profile/models.py
```diff
@@ -0,0 +1,32 @@
...
+            obj=self._username_email_dict(),
+            salt=PROFILE_SALT)
+        return key
+
+    def validate_key(self, activation_key):
+        try:
+            return signing.loads(
+                activation_key,
+                salt=PROFILE_SALT,
+                max_age=timedelta(days=settings.REGISTRATION_VERIFICATION_TIMEOUT_DAYS)
```
das-g: Consider putting the timedelta into the settting instead of an integer. Then the one assigning the setting can choose the granularity.

osmaxx/profile/models.py
```diff
@@ -0,0 +1,32 @@
...
+            )
+        except signing.BadSignature:
+            return None
```
das-g: Why not let the callers of `validate_key()` catch the exception? That'd be more self-explanatory than checking for `None`.

osmaxx/profile/models.py
hixi: I did consider that, and decided against it. Days are quite common, and we're the only ones using this, so why make another setting that needs a description of what it does (or a very very long name to understand, ie `REGISTRATION_VERIFICATION_TIMEOUT_TIMEDELTA` doesn't cut it, because that is just much less clear than `...TIMEOUT_DAYS`). 

I'd rather go in the opposite direction: remove this as a setting altogether and have this inside the code as a constant.

osmaxx/profile/models.py
hixi: Catching the exception at the place closest to where the call causing that exception is IMHO a [good solution](https://github.com/ubernostrum/django-registration/blob/3247d77827bf34beeb11ef49004ad89ec02686fe/registration/backends/hmac/views.py#L124).

osmaxx/profile/views.py
```diff
@@ -1,20 +1,148 @@
...
+        to_email = profile.unverified_email
+        if to_email:
+            cache.set(profile.associated_user.id, 'dummy value', timeout=self.RATE_LIMIT_SECONDS)
+            user_administrator_email = settings.OSMAXX['ACCOUNT_MANAGER_EMAIL']
+            token = profile.activation_key()
+            subject = _('Verify your email address')
+            message = _('In order to verify your email-address click the following link:\n'
+                        '{}?token={}'
+                        '\n'
+                        'If it wasn\'t you, just ignore this email.\n'
```
das-g: Email body could be put into a template file.

osmaxx/profile/views.py
```diff
@@ -1,20 +1,148 @@
...
+                        .format(self.request.build_absolute_uri(reverse('profile:activation')), urlencode(token)))
+            send_mail(
+                subject=subject,
+                message=message,
+                from_email=user_administrator_email,
+                recipient_list=[to_email],
+            )
+            messages.add_message(
+                self.request, messages.INFO, _('To activate your email, click the link in the confirmation email.')
+            )
```
das-g: :+1:

osmaxx/profile/views.py
hixi: putting both in a template would be a good idea, it's much more customizable. Good idea :+1:

osmaxx/profile/views.py
```diff
@@ -1,20 +1,148 @@
...
     template_name = 'profile/profile_edit.html'
 
+    def get(self, *args, **kwargs):
+        profile = self._get_or_create_profile()
+        user = self.request.user
+        if not (user.email or user.profile.unverified_email):
+            messages.add_message(
+                self.request,
+                messages.WARNING,
+                _('You have not set an email. You must set a valid email to use OSMaxx.')
```
das-g: `s/email/email address/g`

osmaxx/profile/forms.py
hixi: Why, is there anything else than an address? Or is it not clear enough without the address?

hixi: After a short search, I'm now convinced, that `email address` is about as common as `email`, but `email address` is actually the correct way of writing it.

das-g: > Why, is there anything else than an address?

Yes, email messages.

> Or is it not clear enough without the address?

It's clear from context that the address is meant. But IMHO "email" without further specification formally only refers to email messages (and maybe the collection of protocols for sending them), even though it's common to colloquially call email addresses just "email", too.

osmaxx/profile/views.py
```diff
@@ -1,20 +1,148 @@
...
+        else:
+            self._ensure_profile_has_email(profile, user)
+        user.refresh_from_db()
+        return super().get(*args, **kwargs)
+
+    def is_new_user(self):
+        user = self.request.user
+        return not (
+            user.profile.unverified_email or user.groups.filter(name=settings.OSMAXX_FRONTEND_USER_GROUP).exists()
+        )
```
das-g: Extracting `user.profile.unverified_email is not None` and `user.groups.filter(name=settings.OSMAXX_FRONTEND_USER_GROUP).exists()` to local variables would allow to indicate what these intermediate values imply without the need for comments:
```python
    def is_new_user(self):
        user = self.request.user
        profile_already_existed = user.profile.unverified_email is not None
        user_already_had_access = user.groups.filter(name=settings.OSMAXX_FRONTEND_USER_GROUP).exists()
        return not (profile_already_existed or user_already_had_access)
```
eliminating the double negation:
```python
    def is_new_user(self):
        user = self.request.user
        profile_is_new = user.profile.unverified_email is None
        user_already_had_access = user.groups.filter(name=settings.OSMAXX_FRONTEND_USER_GROUP).exists()
        return profile_is_new and not user_already_had_access
```

hixi: :+1:

osmaxx/profile/views.py
```diff
@@ -1,20 +1,151 @@
...
     def get_object(self, queryset=None):
-        return self.model.objects.get(pk=self.request.user.id)
+        return self._get_or_create_profile()
 
     def get_success_url(self):
-        messages.add_message(self.request, messages.SUCCESS, 'Profile successfully updated')
+        messages.add_message(self.request, messages.SUCCESS, _('Profile successfully updated'))
         return reverse('profile:edit_view')
+
+    def _get_or_create_profile(self):
```
das-g: consider using https://docs.djangoproject.com/en/1.10/ref/models/querysets/#get-or-create instead

osmaxx/profile/views.py
```diff
@@ -1,20 +1,141 @@
...
+        else:
+            messages.add_message(self.request, messages.ERROR, self.error_msg)
+        return redirect(reverse('profile:edit_view'))
+
+    def _set_group(self, user):
+        if settings.REGISTRATION_OPEN and not self._user_in_frontend_group(user):
+            self._grant_user_access(user)
+
+    def _user_in_frontend_group(self, user):
+        return user.groups.filter(name=settings.OSMAXX_FRONTEND_USER_GROUP).exists()
```
das-g: Could this be turned into a (still private) module-level function, so that it could be used at [`ProfileView.is_new_user()`](https://github.com/geometalab/osmaxx/pull/670/files#diff-c95bcec7c46780ed79611fa9239b3ef1R71), too?

hixi: :+1:

tests/profile/test_activation_view.py
```diff
@@ -0,0 +1,48 @@
...
+
+def test_activation_redirects(authenticated_client, valid_profile):
+    response = authenticated_client.get(reverse('profile:activation'))
+    assert response.status_code == 302
+    response = authenticated_client.get(reverse('profile:activation'), data={'token': valid_profile.activation_key()})
+    assert response.status_code == 302
+    response = authenticated_client.get(reverse('profile:activation'), data={'token': valid_profile.activation_key() + 'm'})
+    assert response.status_code == 302
+    response = authenticated_client.get(reverse('profile:activation'), data={'something_else': 'data'})
+    assert response.status_code == 302
```
das-g: put the four different kwargs into a parametrized fixture

hixi: good idea, forgot about that feature.

tests/profile/test_profile_view.py
```diff
@@ -0,0 +1,91 @@
+import pytest
+from unittest.mock import patch
+
+from django.core.urlresolvers import reverse
+
+
+def test_profile_page_redirects_when_called_without_user(client):
+    response = client.get(reverse('profile:edit_view'))
+    assert response.status_code == 302
+    assert response.url == '/accounts/login/?next=/profile/edit/'
```
das-g: :+1:

PR #672, opened 2016-08-19
make targets for easier use

hixi:
Make it easier for new developers and returning to the project to remember the command sequence for docker-compose.

Reviewed by:
- [x] @das-g

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -142,11 +142,10 @@ up_local_db:
 
 .PHONY: up_local
 up_local: up_local_db
-	${COMPOSE} up -d ${LOCAL_RUN_ONCE_SERVICES}
+	${COMPOSE} up -d ${LOCAL_APPLICATION_STACK}
 
 .PHONY: up_local_all
 up_local_all: build_local up_local_db up_local_run_once up_local
-	${COMPOSE} up -d ${LOCAL_RUN_ONCE_SERVICES}
 
 .PHONY: build_local
 build_local:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -125,3 +125,29 @@ down-pg_translit:
 clean: down-redis down-pg_translit down-pg
 	find . -name __pycache__ -exec rm -rf {} +
 	find . -name "*.pyc" -exec rm -rf {} +
+
+LOCAL_RUN_ONCE_SERVICES := osmboundaries_importer osm_importer
+LOCAL_DB_SERVICES := frontenddatabase mediatordatabase world-database osmboundaries-database
+LOCAL_APPLICATION_STACK := nginx frontend mediator worker worker-exclusive conversionserviceredis
+LOCAL_DEPLOY_VERSION := latest
```
hixi: This should be `latest`, otherwise `build --pull` fails in some cases (compared to `git describe --dirty` as used in our aliases).

das-g:
`python3 -m this | grep preferably`

Should these targets replace [`activate_local_development`](/home/das-g/hsr/dev/osmaxx/osmaxx/activate_local_development) (and the latter therefore be removed in this PR)?

hixi:
I don't think this is to replace activate_local_development, it merely helps to understand why that one might be around in the first place. I also don't think this violates the `one obvious way`, because you can't do the same as with `activate_local_development`, it is just a helper to get one started more quickly.

PR #681, opened 2016-08-19
also allow deletion of exports that are in an undefined state but overdue

### LINKED ISSUE #678: Allow deletion of non-finished exports
Some exports are never finished; Therefore I propose to make the delete option available if it has been more than two days (or whatever we is set on the timeout setting).

hixi:
Reviewed by:
- [x] @das-g 

Closes #678

### DIFF
--- a/osmaxx/excerptexport/models/export.py
+++ b/osmaxx/excerptexport/models/export.py
@@ -163,6 +163,10 @@ class Export(TimeStampModelMixin, models.Model):
         return self.status in self.FINAL_STATUSES
 
     @property
+    def can_be_deleted(self):
+        return self.is_status_final or self.update_overdue
+
+    @property
     def css_status_class(self):
         """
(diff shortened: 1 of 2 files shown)


osmaxx/excerptexport/models/export.py
```diff
@@ -163,6 +163,10 @@ def is_status_final(self):
         return self.status in self.FINAL_STATUSES
 
     @property
+    def can_be_deleted(self):
+        return self.is_status_final or self.update_overdue
```
das-g: :+1: for having logic in model property or method.

Alternative name suggestion: `is_deletable`. (Current name is fine, too, though.)

hixi: I had this name first, but I disliked the PyCharm not knowing and underlining it every time; it even proposed `is_delete_able` as alternative ^^

PR #683, opened 2016-08-24
add excerpt management

### LINKED ISSUE #664: Delete own excerpt boundaries
Let the user delete his own excerpts (not the extracts, but the definitions of the excerpt), maybe on the user profile page and a link on the existing excerpt selection page.

hixi:
Reviewed by:
- [x] @das-g 

Closes #664

### DIFF
--- a/osmaxx/excerptexport/models/excerpt.py
+++ b/osmaxx/excerptexport/models/excerpt.py
@@ -87,7 +87,11 @@ class Excerpt(models.Model):
 
     @property
     def has_running_exports(self):
-        return any(not i.can_be_deleted for es in self.extraction_orders.all() for i in es.exports.all())
+        return any(
+            not export.can_be_deleted
+            for extraction_order in self.extraction_orders.all()
+            for export in extraction_order.exports.all()
+        )
 
     def attached_export_count(self, user):
         return self.extraction_orders.filter(orderer=user).aggregate(Count('exports'))['exports__count']
--- a/osmaxx/excerptexport/models/excerpt.py
+++ b/osmaxx/excerptexport/models/excerpt.py
@@ -88,7 +88,7 @@ class Excerpt(models.Model):
     @property
     def has_running_exports(self):
         return any(
-            not export.can_be_deleted
+            export.is_running
             for extraction_order in self.extraction_orders.all()
             for export in extraction_order.exports.all()
         )
--- a/osmaxx/excerptexport/models/excerpt.py
+++ b/osmaxx/excerptexport/models/excerpt.py
@@ -87,7 +87,7 @@ class Excerpt(models.Model):
 
     @property
     def has_running_exports(self):
-        return any([not i.can_be_deleted for es in self.extraction_orders.all() for i in es.exports.all()])
+        return any(not i.can_be_deleted for es in self.extraction_orders.all() for i in es.exports.all())
 
     def attached_export_count(self, user):
         return self.extraction_orders.filter(orderer=user).aggregate(Count('exports'))['exports__count']
--- a/osmaxx/excerptexport/models/excerpt.py
+++ b/osmaxx/excerptexport/models/excerpt.py
@@ -2,6 +2,7 @@ from django.contrib.auth.models import User
 from django.contrib.gis import geos
 from django.contrib.gis.db import models
 from django.core.cache import cache
+from django.db.models import Count
 from django.utils.translation import gettext_lazy as _
 
 from osmaxx.utils.geometry_buffer_helper import with_metric_buffer
@@ -83,6 +84,13 @@ class Excerpt(models.Model):
     @property
     def extent(self):
         return self.bounding_geometry.extent
+
+    @property
+    def has_running_exports(self):
+        return any([not i.can_be_deleted for es in self.extraction_orders.all() for i in es.exports.all()])
+
+    def attached_export_count(self, user):
+        return self.extraction_orders.filter(orderer=user).aggregate(Count('exports'))['exports__count']
 
     def __str__(self):
         return self.name
--- a/osmaxx/excerptexport/models/export.py
+++ b/osmaxx/excerptexport/models/export.py
@@ -92,7 +92,7 @@ class Export(TimeStampModelMixin, models.Model):
 
     @property
     def update_is_overdue(self):
-        if self.status in self.FINAL_STATUSES:
+        if self.is_status_final:
             return False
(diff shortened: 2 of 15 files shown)


hixi:
@das-g we probably should take a look before merging on where to put the link to the management page.

osmaxx/excerptexport/models/excerpt.py
```diff
@@ -84,6 +85,13 @@ def color(self):
     def extent(self):
         return self.bounding_geometry.extent
 
+    @property
+    def has_running_exports(self):
+        return any([not i.can_be_deleted for es in self.extraction_orders.all() for i in es.exports.all()])
```
das-g: Please use variable name `i` only for integer-valued counters, integer-valued indices etc.

das-g: You can use a generator expression instead of the list comprehension inside `any`. (I.e., leave out the `[` and `]`)

das-g: It seems illogical to have `has_running_exports` (something that can be decided objectively) depend on `Export.can_be_deleted` (the rules for which we've decided subjectively). Also, if the rules of the latter ever change, the results of this new property will become wrong (unless we remember to adapt it).

Suggestion: Rename `Export.can_be_deleted` and alias it as `Export.can_be_deleted`.

osmaxx/excerptexport/models/export.py
```diff
@@ -164,7 +164,7 @@ def is_status_final(self):
 
     @property
     def can_be_deleted(self):
-        return self.is_status_final or self.update_overdue
+        return self.is_status_final or self.update_overdue()
```
das-g: should we make `update_overdue` a property, too?

osmaxx/excerptexport/templates/excerptexport/excerpt_confirm_delete.html
```diff
@@ -0,0 +1,54 @@
...
+                                        </li>
+                                    {% endfor %}
+                                </ul>
+                            </li>
+                        {% empty %}
+                            <li>{% trans 'No Extraction Orders attached.' %}</li>
+                        {% endfor %}
+                    {% else %}
+                        <li>{% trans 'No Extraction Orders attached.' %}</li>
+                    {% endif %}
```
das-g: Why the `{% if %}`-`{% else %}` **and** the `{% empty %}`?

For a non-public excerpt, can `export_count` be `0` without `excerpt.extraction_orders.all` being empty?

osmaxx/excerptexport/templates/excerptexport/excerpt_confirm_delete.html
```diff
@@ -0,0 +1,54 @@
...
+                                            {% capture_as finished_date %}{% firstof export.finished_at|date:"Y-m-d H:i:s" export.updated_at|date:"Y-m-d H:i:s" %}{% end_capture_as %}
+                                            {% capture_as state_display %}{{ export.status }} at {{ finished_date }}{% end_capture_as %}
+                                            {% capture_as available_until %}{{ export.result_file_available_until|date:"Y-m-d H:i:s" }}{% end_capture_as %}
+                                            {% if export.status == export.FINISHED and export.result_file_available_until %}
+                                                {% blocktrans %}<span title="{{ state_display }}">available until {{ available_until }}</span>{% endblocktrans %} /
+                                            {% else %}
+                                                {{ state_display }}
+                                            {% endif %}
+                                            {% if export.output_file.file %}({{ export.output_file.get_filename_display }}){% endif %}
+                                        </li>
```
das-g: I guess we use similar HTML representation of exports elsewhere. Candidate for a partial?

osmaxx/excerptexport/views.py
```diff
@@ -225,3 +225,36 @@ def _social_identification_description(user):
...
+
+    def get_success_url(self):
+        return reverse('excerptexport:manage_own_excerpts')
+
+    def delete(self, request, *args, **kwargs):
+        excerpt = self.get_object()
+        if self.request.user != excerpt.owner:
+            raise GenericViewError("User doesn't match the excerpt's owner.")
+        if excerpt.is_public:
+            raise GenericViewError("No self-defined public excerpts can be deleted.")
```
das-g: Should we also check whether the excerpt is referenced by any extraction orders not owned by the user? Or should users be allowed to delete their own previously public excerpts that have been set private e.g. via the admin interface?

osmaxx/excerptexport/templates/excerptexport/excerpt_confirm_delete.html
hixi: This is a bit messy, therefore a bit hard(er) to grasp, and not logical at all: curently, extraction_orders do exist even after they have been deleted by the user, so the count can be >0 and still have no exports attached.

osmaxx/excerptexport/models/export.py
hixi: I think it would be a good idea, but I'm not happy with the name:
`update_is_overdue` or `is_update_overdue`?

osmaxx/excerptexport/models/excerpt.py
hixi: `i` will be corrected.

generator expression is a good idea, I really should use them more often.

Makes sense, the expression got a bit out of hands, when using the `can_be_deleted` from exports, I'm not fan of this implementation in general. Do you have any smart suggestions (besides renaming)?

osmaxx/excerptexport/templates/excerptexport/excerpt_confirm_delete.html
hixi: If by partial you mean [inclusion](https://docs.djangoproject.com/en/1.10/ref/templates/builtins/#include) I concur :+1:

osmaxx/excerptexport/models/export.py
das-g: Good idea to include `is` in the name (whether this stays a method or becomes a property). I'm impartial between `update_is_overdue` and `is_update_overdue`.

osmaxx/excerptexport/views.py
hixi: I had something similar in mind, but thought that very edgy-case is not important enough. Maybe it actual *is* a good idea to catch this case...

osmaxx/excerptexport/templates/excerptexport/excerpt_confirm_delete.html
das-g: Yes, I meant a partial template to be included here.

das-g:
:+1: good to merge (as soon as tests have passed)

PR #706, opened 2016-10-04
Feature/#705 qgis project in export zip

### LINKED ISSUE #705: include QGIS project file in export download ZIP
With the references adapted to point to the respective data file for the export.

das-g:
closes #705

Additionally to what #705 asks for, this sets the map canvas of the generated QGIS project to the excerpt extent.

### Reviewed by
- [ ] @hixi
- [ ] @sfkeller

To avoid the copy-paste noise, you might want to view the [diff from f39abd5 to `feature/#705_QGIS_project_in_export_ZIP`](https://github.com/geometalab/osmaxx/compare/feature/%23705_QGIS_project_in_export_ZIP_preparation...feature/%23705_QGIS_project_in_export_ZIP?expand=1) rather than from `develop` to `feature/#705_QGIS_project_in_export_ZIP`.

- [ ] I'm not sure whether refactoring 3a4d4c5 actually improves anything. Should I revert that change?


osmaxx/conversion_api/formats.py
```diff
@@ -54,6 +75,8 @@ def detail_level_available(self):
         archive_file_name_identifier='Shapefile',
         abbreviations=[],
         is_white_box=True,
+        layer_filename_extension='.shp',
+        attribute_value_encoding='ISO-8859-1',
```
das-g: This (a.k.a. `latin1`) seems to be the default encoding of Shapefiles if no `*.cpg` file is present and the LDID/codepage setting from the `*.dbf` file doesn't specify anything. It's probably also what OGR creates (when writing Shapefiles) when not told otherwise.

However, `latin1` will not be able to properly encode some values that will occur in the data (e.g. non-western names in respective language's native script). @sfkeller Thus, should we instead force OGR to create Shapefiles with e.g. UTF-8 encoded attribute values in the `*.dbf` files along with appropriate `*.cpg` files specifying that encoding?

das-g: Deferred to #708

PR #717, opened 2016-10-05
trim leading and following whitespace

### LINKED ISSUE #712: text change: delete all items ..." ..."
Remove leading space after first ".

hixi:
closes #712


osmaxx/excerptexport/static/excerptexport/scripts/add_delete_button_to_exports.js
```diff
@@ -35,7 +35,7 @@
                 group_title.attr("id", "delete-export-all");
                 group_title.prepend('<span class="glyphicon glyphicon-remove-circle hand-cursor" aria-hidden="true"></span> ');
```
das-g: I think the space comes from here. So wouldn't it be better to simply evaluate `group_title.text()` already before doing this `.prepend()` instead of trimming afterwards?

PR #729, opened 2016-10-19
Fix XSS-Protection Header sent twice

hixi:
this header is already set in the production nginx, this way it wil be set twice; therefore removed this line


hixi:
Django does set [this header](https://docs.djangoproject.com/en/1.10/ref/settings/#secure-browser-xss-filter) itself with [our configuration](https://github.com/geometalab/osmaxx/blob/0d067069bb733ff32c3ab4a5eba36a059b4edb5b/docker-compose-common.yml#L7), that's the reason for the duplication.

PR #734, opened 2016-10-26
switch to pbf cutter instead of using a world database

hixi:
Reviewed by:
- [x] @das-g 

switch to pbf cutting instead of using the world database.

To be done:
- [x] pbf updater
- [x] remove dependency on the world database container
- [x] tests


osm_pbf_updater/pbf_updater.py
```diff
@@ -0,0 +1,64 @@
...
+    if osmupdate_extra_params is not None:
+        update_comand += osmupdate_extra_params.split()
+    update_comand += [PLANET_LATEST, PLANET_LATEST_ON_UPDATE]
+    subprocess.check_call(NICE + update_comand)
+    shutil.move(PLANET_LATEST_ON_UPDATE, PLANET_LATEST)
+
+
+def run(*, sleep_seconds=10, osmupdate_extra_params=None):
+    while True:
+        # no loss, whenever we start the container up, we download the newest pbf and start updating it.
```
das-g: I'm not sure I understand what this comment wants to tell me. "No loss" of what? Compared to what?

osm_pbf_updater/pbf_updater.py
```diff
@@ -0,0 +1,64 @@
...
+        initial_download(planet_url())
+        # start updating immediately
+        update(osmupdate_extra_params)
+        print("update done, sleeping for {} seconds".format(sleep_seconds))
```
das-g: consider to [`log`](https://docs.python.org/howto/logging.html) instead of `print`ing.

osm_pbf_updater/pbf_updater.py
```diff
@@ -0,0 +1,64 @@
...
+
+if __name__ == '__main__':
+    import argparse
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-t", "--wait-seconds",
+                        type=int, default=120,
+                        help="wait this amount of seconds between updates")
+    args = parser.parse_args()
+    sleep_seconds = args.wait_seconds
+    update_extra_params = os.environ.get('osmupdate_extra_params', None)
```
das-g: Why not use an empty string as the fallback? That'd simplify the code of `update`.

osm_pbf_updater/pbf_updater.py
hixi: No, logging inside the container does not output that to the default log, and one would need to configure logging to output the log to print it to the tty/console. If I look inside the containers log, I'd like to see this message.

osm_pbf_updater/pbf_updater.py
hixi: Good idea :+1:

PR #736, opened 2016-11-07
fix #708: Encode attribute values as UTF-8 in Shapefiles, too

### LINKED ISSUE #708: Encode attribute values as UTF-8 in Shapefiles, too
See https://github.com/geometalab/osmaxx/pull/706#pullrequestreview-2735568

das-g:
fixes #708

Encode attribute values as UTF-8 in Shapefiles, too and let QGIS treat the values as UTF-8 encoded.
### Reviewed by
- [x] @hixi
- [ ] @sfkeller

### DIFF
--- a/osmaxx/conversion/converters/converter_gis/extract/db_to_format/extract.py
+++ b/osmaxx/conversion/converters/converter_gis/extract/db_to_format/extract.py
@@ -18,7 +18,7 @@ FORMATS = {
     SHAPEFILE: {
         'ogr_name': 'Esri Shapefile',
         'extension': '.shp',
-        'extraction_options': [],
+        'extraction_options': ['-lco', 'ENCODING=UTF-8'],
     },
     SPATIALITE: {
         'ogr_name': 'SQLite',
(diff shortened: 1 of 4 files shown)


das-g:
Should we remove the attribute value encoding parametrization (introduced in https://github.com/geometalab/osmaxx/pull/706/commits/ba6673f475afec280ad5e2453fb957f9e955c986) again, now that all formats use UTF-8? (If so, let's merge 02c7b8f2b9df3e133274ccc7d369acb8ca609e31 into this branch before merging this PR.)

hixi:
why would we want to hardcode this? Does that have any negative effects?

das-g:
> why would we want to hardcode this?

Because now (hopefully) YAGNI anymore.

> Does that have any negative effects leaving it this way?

No severe ones, just

- slightly more code
- more stuff to search-replace when converting new revisions of the QGIS project file to templates

hixi:
The last point, replacing it for every change does make sense to reduce that work, IMHO. Let's merge 02c7b8f :+1:

PR #746, opened 2016-11-16
Feature/geometalab/osmaxx-symbology#99 scale dependent visibility in qgis

das-g:
closes geometalab/osmaxx-symbology#99

# TODO:
* [ ] create new sample export


hixi:
@das-g I just added a new example export, if you think this is ready please merge.

PR #761, opened 2016-11-24
Feature/fix estimation failure

hixi:
only drawback: no estimation of the pbf size in case of failure.

@das-g please review


tests/conversion/management/commands/result_harvester_test.py
```diff
@@ -0,0 +1,25 @@
...
+    from osmaxx.conversion.management.commands.result_harvester import add_meta_data_to_job
+    conversion_job = MagicMock()
+    conversion_job.unzipped_result_size = None
+    conversion_job.extraction_duration = None
+    conversion_job.estimated_pbf_size = None
+    conversion_job.parametrization.clipping_area.clipping_multi_polygon.extent = -70, 70, 80, -10
+    rq_job = Mock()
+    rq_job.meta = {
+        'unzipped_result_size': 0,
+        'duration': 0,
```
das-g: If you've bothered to test these (see assertions below), maybe use values more unique than `0` ...

hixi: is `42` better? ;-)

PR #766, opened 2016-11-28
ensure a pbf file is being downloaded in case it doesn't exist

hixi:
download the pbf file in case it doesn't exist just yet.

### DIFF
--- a/osm_pbf_updater/pbf_updater.py
+++ b/osm_pbf_updater/pbf_updater.py
@@ -45,6 +45,11 @@ def _is_night_time(now):
 
 def run(*, sleep_seconds=10, osmupdate_extra_params):
     last_full_download_time = datetime.datetime.min
+
+    if not os.path.exists(PLANET_LATEST):
+        full_download(planet_url())
+        last_full_download_time = datetime.datetime.now()
+
     while True:
         now = datetime.datetime.now()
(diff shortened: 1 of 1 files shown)


osm_pbf_updater/pbf_updater.py
```diff
@@ -45,6 +45,11 @@ def _is_night_time(now):
...
 def run(*, sleep_seconds=10, osmupdate_extra_params):
     last_full_download_time = datetime.datetime.min
+
+    if not os.path.exists(PLANET_LATEST):
+        full_download(planet_url())
+        last_full_download_time = datetime.datetime.now()
+
     while True:
         now = datetime.datetime.now()
         if now - last_full_download_time > datetime.timedelta(weeks=1) and _is_night_time(now):
```
das-g: How about including that here:
```python
if not os.path.exists(PLANET_LATEST) or (
        now - last_full_download_time > datetime.timedelta(weeks=1) and _is_night_time(now)):
    full_download(planet_url())
    last_full_download_time = datetime.datetime.now()
```

hixi: I kind of like to have this as a `quasi` precondition before the loop starts, but I have no objection to change it to your suggestion. Should I change it or leave it?

PR #767, opened 2016-11-28
don't crash if pbf file is already up to date

hixi:
don't crash if pbf file is already up to date

### DIFF
--- a/osm_pbf_updater/pbf_updater.py
+++ b/osm_pbf_updater/pbf_updater.py
@@ -35,7 +35,25 @@ def full_download(complete_planet_mirror_url):
 
 def update(osmupdate_extra_params):
     update_comand = ["osmupdate", "-v"] + osmupdate_extra_params.split() + [PLANET_LATEST, PLANET_LATEST_ON_UPDATE]
-    subprocess.check_call(NICE + update_comand)
+    try:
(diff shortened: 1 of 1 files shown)


osm_pbf_updater/pbf_updater.py
```diff
@@ -35,7 +35,25 @@ def full_download(complete_planet_mirror_url):
...
+        #     if(!file_exists(master_cachefile_name)) {
+        #       if(old_file==NULL)
+        #         PINFO("There is no changefile since this timestamp.")
+        #       else
+        #         PINFO("Your OSM file is already up-to-date.")
+        # return 21;
+        if e.returncode == 21:
+            pass
+        else:
+            raise
```
das-g: Can of course be written more compactly as
```python
except subprocess.CalledProcessError as e:
    if e.returncode != 21:
        raise
```
but for the usage here, I kinda like the variant with `pass`.

hixi: I'm not convinced that my (late night) version is better, but it certainly states the intent better (by coincidence).

das-g: > but it certainly states the intent better

That's why I like it :smile:

PR #772, opened 2016-12-13
Feature/rework landing page emblems

das-g:
## before:
screenshot of frontpage before change
## after:
screenshot of frontpage after change


sfkeller:
The new emblems graphics is OK for me.

I just see that the buttons still don't have the same text as the menu. I'd prefer a hover text over the two emblems, and that the emblems are "clickable".

Even more, I'm wondering, if the buttons below the two emblems can't be completely omitted?

das-g:
> I just see that the buttons still don't have the same text as the menu. I'd prefer a hover text over the two emblems, [...]

Good point. Let's handle that in a separate pull request, though: #773 (hovertext only over buttons, but we I can change that)

> I'd prefer a hover text over the two emblems, and that the emblems are "clickable".
> 
> Even more, I'm wondering, if the buttons below the two emblems can't be completely omitted?

Maybe the caption and the emblems should be together on a button, to indicate that the whole thing is clickable?

PR #778, opened 2017-01-11
Feature/admin overhaul

hixi:
Adds support for listing the user's exports and its dependent data in the admin.

### DIFF
--- a/osmaxx/excerptexport/admin.py
+++ b/osmaxx/excerptexport/admin.py
@@ -35,8 +35,8 @@ class OutputFileAdmin(admin.ModelAdmin):
 
 @admin.register(Export)
 class ExportAdmin(admin.ModelAdmin):
-    list_display = ('id', 'file_format', 'conversion_service_job_id', 'status', 'finished_at')
+    list_display = ('id', 'file_format', 'conversion_service_job_id', 'status', 'created_at', 'finished_at')
     list_display_links = ('id', )
-    list_filter = ['finished_at', 'created_at', 'updated_at']
+    list_filter = ['status', 'finished_at', 'created_at', 'updated_at']
     fields = ('file_format', 'conversion_service_job_id', 'status', 'finished_at', 'created_at', 'updated_at')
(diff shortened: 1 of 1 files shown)


osmaxx/excerptexport/admin.py
```diff
@@ -35,6 +53,36 @@ class OutputFileAdmin(admin.ModelAdmin):
...
+    readonly_fields = ('finished_at', 'created_at', 'updated_at', 'link_to_extraction_order')
+
+    def link_to_extraction_order(self, obj):
+        link = urlresolvers.reverse("admin:excerptexport_extractionorder_change", args=[obj.extraction_order.id])
+        return u'<a href="%s">%s</a>' % (link, obj.extraction_order.excerpt_name)
+    link_to_extraction_order.allow_tags = True
+
+
+User = get_user_model()
+admin.site.unregister(User)
```
das-g: I'd put these two lines directly above
```python
@admin.register(User)
class UserAdmin2(UserAdmin):
    # ...
```
below.

osmaxx/excerptexport/admin.py
```diff
@@ -35,6 +53,36 @@ class OutputFileAdmin(admin.ModelAdmin):
...
+    show_change_link = True
+    extra = 0
+
+    def has_add_permission(self, request):
+        return False
+
+
+@admin.register(User)
+class UserAdmin2(UserAdmin):
+    inlines = [
```
das-g: As `UserAdmin` might declare its own `inlines` in the future, should this be
```python
    inlines += [
        # ...
```
or would we _want_ to override pre-existing `inlines` in that case?

osmaxx/excerptexport/admin.py
hixi: good idea, I missed that one.

osmaxx/excerptexport/admin.py
hixi: wouldn't this result in "unresolved reference"?

PR #784, opened 2017-01-16
Fix/dont pass content type for output file creation

das-g:
Our mistake. Django 1.10 finally makes it break.


das-g:
@hixi can't log in today, but said:
> has aglueged und isch guet so! :+1:

PR #786, opened 2017-01-18
Feature/django 1.10 fix proposal

hixi:


Dockerfile.worker
```diff
@@ -1,4 +1,4 @@
-# this gdal image comes with support for FileGDB
+# this gdal image comes with support for FileGDB and has python3.4 already installed
 FROM geometalab/gdal-debian:2.1.0
```
das-g: consider using `latest` here

Dockerfile.frontend
```diff
@@ -1,7 +1,10 @@
-FROM geometalab/python3-gis
+FROM geometalab/gdal-debian
+
+USER root
```
das-g: Should we switch back to user `py` at the end (or whereever the Dockerfile doesn't need `root` anymore) or does our entrypoint need `root`, too?

PR #793, opened 2017-01-27
Feature/multiple queue fix

hixi:
fixes the multiple queue issue.

### DIFF
--- a/tests/conversion/management/commands/result_harvester_test.py
+++ b/tests/conversion/management/commands/result_harvester_test.py
@@ -70,3 +70,24 @@ def test_add_meta_data_to_job_with_out_of_bounds_exception():
         )
         logger_mock.exception.assert_called_once_with("pbf estimation failed")
         assert conversion_job.estimated_pbf_size is None
+
+
+def multiple_queue_test_parameters():
+    queue = Mock()
+    none_queue = Mock()
+    none_queue.fetch_job.return_value = None
+    return [
+        ([queue, none_queue], queue.fetch_job.return_value),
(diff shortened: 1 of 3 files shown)


tests/conversion/management/commands/result_harvester_test.py
```diff
@@ -70,3 +70,24 @@ def test_add_meta_data_to_job_with_out_of_bounds_exception():
         )
         logger_mock.exception.assert_called_once_with("pbf estimation failed")
         assert conversion_job.estimated_pbf_size is None
+
+
+def multiple_queue_test_parameters():
+    queue = Mock()
+    none_queue = Mock()
+    none_queue.fetch_job.return_value = None
```
das-g: How about more informative names: e.g. `queue_with_all_jobs` and `queue_with_no_jobs`?

das-g: It would be nice to extract `queue` and `none_queue` as fixtures, but py.test [doesn't yet allow this without making all of this much more verbose and very ugly](https://github.com/pytest-dev/pytest/issues/349). :disappointed:

Alex-Bogdanov:
Sorry, when this 2 features will be released?http://doc.pytest.org/en/latest/proposals/parametrize_with_fixtures.html
`pytest.define_combined_fixture()`
` pytest.fixture_request()`

das-g:
What does this question have to do with this pull request?

The [pytest repo](https://github.com/pytest-dev/pytest/) might be a better place to ask this question than the OSMaxx repo.

PR #797, opened 2017-01-30
Feature/refactor rq job deleting function

hixi:
refactoring to remove ugly function and concatenate all queues.


das-g:
much nicer :+1:

PR #798, opened 2017-01-30
Feature/remove unused languages

hixi:
reduced all languages to one, `en`.


release.py
```diff
@@ -62,8 +62,10 @@ def create_data_schema_documentation():
 
 def make_release_specific_changes(release_version):
     execute_steps([
-        "python ./web_frontend/manage.py makemessages -l de_CH -l en -l en_UK -l en_US",
-        "python ./osmaxx_conversion_service/manage.py makemessages -l de_CH -l en -l en_UK -l en_US",
+        "python ./web_frontend/manage.py makemessages -l en",
+        "python ./osmaxx_conversion_service/manage.py makemessages -l en",
```
das-g: Shouldn't these two commands produce one `*.po` file each, thus two in total? [Only one](https://github.com/geometalab/osmaxx/pull/798/files#diff-36e525b0fdeb572e17dd3158ee2245e8) seems to remain. Are they overwriting each other's output?

release.py
```diff
@@ -62,8 +62,10 @@ def create_data_schema_documentation():
...
+        "python ./web_frontend/manage.py compilemessages -l en",
+        "python ./osmaxx_conversion_service/manage.py compilemessages -l en",
```
das-g: Good catch :+1:

:question: Though, same here: Shouldn't these two commands produce one `*.mo` file each, thus two in total? I see [only one](https://github.com/geometalab/osmaxx/pull/798/files#diff-562dd48f7010d93324f62151cc1b2126) being newly tracked. Are they overwriting each other's output?

release.py
hixi: Good catch, they should actually write to the same file, and since they have somewhat the same codebase, this `should` produce essentially the same file content. Unfortunately this is not the case, because each has different apps loaded.

-> I would argue, that this needs some cleanup: Only the translations of the customer facing website should even be considered.

PR #804, opened 2017-02-03
Feature/#167 pbf export format

### LINKED ISSUE #167: Support *.osm.pbf as export format
Allow users to download the 'planet' file (.osm.pbf) of an extract.

hixi:
add support for exporting pbf format

closes #167

### DIFF
--- a/osmaxx/conversion/converters/converter_gis/bootstrap/bootstrap.py
+++ b/osmaxx/conversion/converters/converter_gis/bootstrap/bootstrap.py
@@ -5,7 +5,7 @@ import subprocess
 from osmaxx.conversion.converters.converter_gis.helper.default_postgres import get_default_postgres_wrapper
 from osmaxx.conversion.converters.converter_gis.helper.osm_boundaries_importer import OSMBoundariesImporter
 from osmaxx.conversion.converters import detail_levels
-from osmaxx.conversion.converters.converter_pbf.to_pbf import polyfile_string_to_pbf
+from osmaxx.conversion.converters.converter_pbf.to_pbf import cut_pbf_along_polyfile
 from osmaxx.conversion.converters.detail_levels import DETAIL_LEVEL_TABLES
 from osmaxx.utils import polyfile_helpers
 
@@ -22,7 +22,7 @@ class BootStrapper:
 
     def bootstrap(self):
         self._reset_database()
-        polyfile_string_to_pbf(self.area_polyfile_string, self._pbf_file_path)
+        cut_pbf_along_polyfile(self.area_polyfile_string, self._pbf_file_path)
         self._import_boundaries()
         self._import_pbf()
         self._setup_db_functions()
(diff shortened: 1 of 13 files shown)


osmaxx/conversion/converters/converter_gis/bootstrap/bootstrap.py
```diff
@@ -23,22 +23,26 @@ def cut_area_from_pbf(pbf_result_file_path, extent_polyfile_path):
     subprocess.check_call(command)
 
 
+def polyfile_string_to_pbf(polyfile_string, pbf_out_path):
+    polyfile_path = os.path.join('/tmp', 'polyfile_extent.poly')
```
das-g: Now that the generated `*.poly` file is only used locally (i.e. only inside this function), we could use a [`NamedTemporaryFile`](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile).

osmaxx/conversion/converters/converter_pbf/to_pbf.py
```diff
@@ -0,0 +1,56 @@
...
+        f.write(polyfile_string)
+    cut_area_from_pbf(pbf_out_path, polyfile_path)
+
+
+def produce_pbf(*, out_zip_file_path, area_name, polyfile_string):
+    _start_time = timezone.now()
+
+    with tempfile.TemporaryDirectory() as tmp_dir:
+        out_dir = os.path.join(tmp_dir, 'pbf')
+        os.makedirs(out_dir, exist_ok=True)
```
das-g: `exist_ok=True` should be unnecessary here, as a new temp dir should be empty.

hixi: You just used should twice, so you are not certain either? :grin: 
I followed the principle:
Is it an error, if it already exists? No.
Is it possible, that it already exist? Not very likely.
Does it any harm, leaving it in there? It is a bit confusing, at best.
Do we have the same pattern somewhere else? [Yes, and we do the same](https://github.com/geometalab/osmaxx/blob/develop/osmaxx/conversion/converters/converter_garmin/garmin.py#L65)

Result was: Mimik the code we already have, so that refactoring might actually catch that line as well (if we should check for duplicated code, this is more obvious).

osmaxx/conversion/converters/converter_gis/bootstrap/bootstrap.py
hixi: Good idea :+1:

osmaxx/conversion/converters/converter_gis/bootstrap/bootstrap.py
```diff
@@ -48,7 +52,7 @@ def bootstrap(self):
 
     @property
     def geom(self):
-        return self._extent
+        return polyfile_helpers.parse_poly_string(self.area_polyfile_string)
```
hixi: probably not, but it shouldn't be done in `__init__` since we don't always use it.

das-g: Should we use [`memoize`](https://pypi.python.org/pypi/memoize) for it or implement a simple memoization ourselves? Or just make it a regular method instead of a property, so that the weight of invoking it is better communicated.

PR #805, opened 2017-02-08
Feature/log check output calls

hixi:
log output and failure when check_call is used.

### DIFF
--- a/tests/utils/test_logged_subprocess_wrapper.py
+++ b/tests/utils/test_logged_subprocess_wrapper.py
@@ -1,38 +1,41 @@
 import pytest
+import subprocess
+
+from osmaxx.conversion.converters import utils
 
 
-def test_logged_check_call_accepts_params(mocker):
-    from osmaxx.conversion.converters.utils import logged_check_call
-    subprocess_check_call_mock = mocker.patch('osmaxx.conversion.converters.utils.subprocess.check_call')
+@pytest.fixture
+def subprocess_check_call_mock(mocker):
+    return mocker.patch.object(subprocess, 'check_call')
+
+
+def test_logged_check_call_accepts_params(subprocess_check_call_mock):
     test_call = ['echo', '0']
-    logged_check_call(test_call)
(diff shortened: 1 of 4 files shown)


tests/utils/test_logged_subprocess_wrapper.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def test_logged_check_call_accepts_params(subprocess_check_call_mock):
+    test_call = ['echo', '0']
+    utils.logged_check_call(test_call)
+    subprocess_check_call_mock.assert_called_with(test_call)
+
+    test_call = "echo 0"
+    utils.logged_check_call(test_call, shell=True)
+    subprocess_check_call_mock.assert_called_with(test_call, shell=True)
```
das-g: parametrize this test instead of asserting twice

tests/utils/test_logged_subprocess_wrapper.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def test_logged_check_call_raises():
+    with pytest.raises(subprocess.CalledProcessError) as excinfo:
+        test_call = ['false']
```
das-g: Consider inlining `test_call`.

If you don't, pull it out of the `with` block so that it's clearer where we expect the error to be raised.

tests/utils/test_logged_subprocess_wrapper.py
```diff
@@ -0,0 +1,41 @@
...
+        utils.logged_check_call(test_call)
+    assert excinfo.value.output is None
+    assert excinfo.value.returncode == 1
+
+
+def test_logged_check_call_logs_error_content(mocker, subprocess_check_call_mock):
+    error_output = 'example output that is available'
+    error_return_code = 17
+    command = 'cmd'
```
das-g: consider inlining these two variables and eliminating the `.format` in the assertion (i.e. use a hardcoded string for expected logmessage, with the substitutions already done)

PR #826, opened 2017-03-20
Feature/fix template filter siteabsoluteurl

das-g:


osmaxx/core/templatetags/navigation.py
```diff
@@ -1,19 +1,34 @@
-# Adopted from https://djangosnippets.org/snippets/2783/
+from urllib.parse import urlsplit, urlunsplit
+
+import logging
 from django import template
 from django.conf import settings
+from django.http import HttpRequest
+
+logger = logging.getLogger(__name__)
```
hixi: :+1:

osmaxx/core/templatetags/navigation.py
```diff
@@ -1,19 +1,34 @@
...
 
 register = template.Library()
 
 
 @register.filter
-def siteabsoluteurl(absoluteurl, request):
+def siteabsoluteurl(relative_or_absolute_url: str, request: HttpRequest) -> str:
```
hixi: IMO unnecessary addition, because this function is used in templates only. It doesn't help to understand the code, the parameters are already quite sensibly named without the need of type information. Furthermore we don't use that at any other place in the project, unless we plan to include this pattern in all place in the future, I think we should not start using it.

osmaxx/core/templatetags/navigation.py
```diff
@@ -1,19 +1,34 @@
...
+        An absolute URL, incl. scheme, host, etc.
     """
-    absolute_url = request.build_absolute_uri(absoluteurl)
-    if settings.OSMAXX.get('SECURED_PROXY', False) and not absolute_url.startswith('https'):
-        if absolute_url.startswith('http'):
-            absolute_url = absolute_url[4:]
-        absolute_url = 'https' + absolute_url
+    absolute_url = request.build_absolute_uri(relative_or_absolute_url)
+    if settings.OSMAXX.get('SECURED_PROXY', False):
+        url_components = urlsplit(absolute_url)
```
hixi: much better using stdlib! :+1:

osmaxx/core/templatetags/navigation.py
das-g: > IMO unnecessary addition, because this function is used in templates only. It doesn't help to understand the code, the parameters are already quite sensibly named without the need of type information.

I've added these type hints exactly because when making this change, I wasn't sure of the types and had a look at some other code to figure them out. While the names informed me about the arguments' contents, I didn't know the types. I was tempted to put the type information into the docstring, but then I remembered that we have an even more appropriate place for it in Python &ge; 3.5 (compatible to Python &ge; 3.0).

> Furthermore we don't use that at any other place in the project,

Correct.

> unless we plan to include this pattern in all place in the future, I think we should not start using it.

I beg to differ: I think we should use it where we see a need for it, without making it compulsory. Python type hints were specifically designed so that type-hinted and non-type-hinted code mix well, so that the hints can be used as-needed.

osmaxx/core/templatetags/navigation.py
das-g: Yeah. IIRC, we already considered using `urllib.parse` back when this filter was first implemented, but didn't use it because we didn't want to "introduce unnecessary dependencies". :thinking: Noticing recently that `request.build_absolute_uri()` (which we already used then) uses `urllib.parse`, too, I thought that point was moot. :stuck_out_tongue_closed_eyes:

osmaxx/core/templatetags/navigation.py
hixi: This is not an argument to win or loose, I don't care enough to waste time arguing about this. If it helps you understand this piece of code, I'll accept it as it stands ;-)

PR #828, opened 2017-03-21
Feature/#755 display time zone with datetimes

### LINKED ISSUE #755: display TimeZone time is in
For exports, display timezone (UTC+1 or so) for the displayed time.

das-g:
closes #755

### DIFF
--- a/osmaxx/excerptexport/templates/excerptexport/partials/export_detail_string.html
+++ b/osmaxx/excerptexport/templates/excerptexport/partials/export_detail_string.html
@@ -1,7 +1,7 @@
 {% load i18n capture_as %}{{ export.get_file_format_display }} /
-{% firstof export.finished_at|date:"Y-m-d H:i:s" export.updated_at|date:"Y-m-d H:i:s" as finished_date %}
+{% firstof export.finished_at|date:"Y-m-d H:i:s e" export.updated_at|date:"Y-m-d H:i:s e" as finished_date %}
 {% capture_as state_display %}{{ export.status }} at {{ finished_date }}{% end_capture_as %}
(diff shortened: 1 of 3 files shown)


osmaxx/excerptexport/templates/excerptexport/partials/export_detail_string.html
```diff
@@ -1,7 +1,7 @@
 {% load i18n capture_as %}{{ export.get_file_format_display }} /
-{% firstof export.finished_at|date:"Y-m-d H:i:s" export.updated_at|date:"Y-m-d H:i:s" as finished_date %}
```
hixi: Instead of repeating this format, we could just use it as default setting: https://docs.djangoproject.com/en/1.10/ref/settings/#std:setting-DATETIME_FORMAT
`{{ value|date:"DATETIME_FORMAT" }}` 
Also see https://docs.djangoproject.com/en/1.10/ref/templates/builtins/#date

das-g: good point, will change that

das-g: How do I do that, though? With 0062085dfed062d2b87b09978a2e0f526e3e23f6, the datetimes are displayed as e.g.
> March 21, 2017, 4:15 p.m

Is that because
> [...] predefined formats may vary depending on the current locale.

as noted in the [`date:'...'` filter documentation](https://docs.djangoproject.com/en/1.10/ref/templates/builtins/#date)?

das-g: Ah, [this](https://mounirmesselmeni.github.io/2014/11/06/date-format-in-django-admin/) works. :-)

PR #830, opened 2017-03-24
use released version of osm2pgsql not master

### LINKED ISSUE #829: GIS export of Haiti failing
[`osmaxx/conversion/converters/converter_gis/bootstrap/sql/filter/building/010_building.sql`](https://github.com/geometalab/osmaxx/blob/v3.7.3/osmaxx/conversion/converters/converter_gis/bootstrap/sql/filter/building/010_building.sql) is causing
```
sqlalchemy.exc.DataError: (psycopg2.DataError) Geometry type (MultiLineString) does not match column type (MultiPolygon)
```
https://sentry.njo.ch/geometalab/osmaxxprod/issues/306/

hixi:
Closes #829

### DIFF
--- a/Dockerfile.worker
+++ b/Dockerfile.worker
@@ -104,10 +104,12 @@ RUN apt-get clean && apt-get update && \
 WORKDIR /root/osm2pgsql
 
 # OSM2PGSQL
+ENV OSM2PGSQL_VERSION=0.92.0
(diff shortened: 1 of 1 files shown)


das-g:
(test on Travis aborted, because only a dockerfile was changed and we don't test these)

Issue #831, opened 2017-03-27
slashes in name of export lead to failure


hixi:
When slashes are used in the export name, this fails, since the name is used as-is and that directory doesn't exists.

Proposal: [slugify](https://docs.djangoproject.com/en/1.10/ref/utils/#django.utils.text.slugify) the name

das-g:
reproduced: https://sentry.njo.ch/geometalab/osmaxxstaging/issues/310/

PR #837, opened 2017-04-07
Feature/#831 slugify pbf name

das-g:
files #831

### DIFF
--- a/osmaxx/conversion/converters/converter.py
+++ b/osmaxx/conversion/converters/converter.py
@@ -60,7 +60,7 @@ class Conversion(object):
     def _create_pbf(self):
         produce_pbf(
             out_zip_file_path=self._output_zip_file_path,
-            area_name=self._area_name,
+            filename_prefix=self._area_name,
             polyfile_string=self._polyfile_string,
         )
 
--- a/osmaxx/conversion/converters/converter.py
+++ b/osmaxx/conversion/converters/converter.py
@@ -60,7 +60,7 @@ class Conversion(object):
     def _create_pbf(self):
         produce_pbf(
             out_zip_file_path=self._output_zip_file_path,
-            filename_prefix=self._area_name,
+            filename_prefix=self._name_prefix,
             polyfile_string=self._polyfile_string,
         )
 
--- a/osmaxx/conversion/converters/converter_pbf/to_pbf.py
+++ b/osmaxx/conversion/converters/converter_pbf/to_pbf.py
@@ -31,13 +31,13 @@ def cut_pbf_along_polyfile(polyfile_string, pbf_out_path):
         cut_area_from_pbf(pbf_out_path, polyfile.name)
 
 
-def produce_pbf(*, out_zip_file_path, area_name, polyfile_string):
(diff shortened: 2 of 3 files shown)


hixi:
`_name_prefix` isn't sluggified in this pull request, is this being handled in a separate pull-request or is it just missing here?

das-g:
> `_name_prefix` isn't sluggified in this pull request, is this being handled in a separate pull-request or is it just missing here?

Neither: If I'm not mistaken, it was already sufficiently slugified before this PR. This PR just makes the PBF export use it, instead of the bare `area_name`.

We already expect `_name_prefix` to be, well, a valid partial file name. (Otherwise, exporting excerpts with a `/` in their name would fail for other formats, too.) And indeed, [`produce_pbf(...)`](https://github.com/geometalab/osmaxx/blob/c32f9dd8817db9ff6abf3027b4d028dc1c3d1adb/osmaxx/conversion/converters/converter_pbf/to_pbf.py#L34) seems to be only called from [`Conversion._create_pbf(...)`](https://github.com/geometalab/osmaxx/blob/c32f9dd8817db9ff6abf3027b4d028dc1c3d1adb/osmaxx/conversion/converters/converter.py#L60), which (as [`Conversion._conversion_process(...)`](https://github.com/geometalab/osmaxx/blob/c32f9dd8817db9ff6abf3027b4d028dc1c3d1adb/osmaxx/conversion/converters/converter.py#L36) is called only from [`Conversion.start_format_extraction(...)`](https://github.com/geometalab/osmaxx/blob/c32f9dd8817db9ff6abf3027b4d028dc1c3d1adb/osmaxx/conversion/converters/converter.py#L38) which (outside of tests) seems to only be called from [`converter.convert()`](https://github.com/geometalab/osmaxx/blob/c32f9dd8817db9ff6abf3027b4d028dc1c3d1adb/osmaxx/conversion/converters/converter.py#L68), which [initializes the conversion with its `filename_prefix` argument](https://github.com/geometalab/osmaxx/blob/c32f9dd8817db9ff6abf3027b4d028dc1c3d1adb/osmaxx/conversion/converters/converter.py#L77).

`converter.convert()` itself is (outside of tests) only called [by itself (with the same arguments, except for `use_worker`)](https://github.com/geometalab/osmaxx/blob/c32f9dd8817db9ff6abf3027b4d028dc1c3d1adb/osmaxx/conversion/converters/converter.py#L85) and by [`Job.start_conversion(...)`](https://github.com/geometalab/osmaxx/blob/c32f9dd8817db9ff6abf3027b4d028dc1c3d1adb/osmaxx/conversion/models.py#L60) it seems, [the latter passing `Job._filename_prefix()` as the `filename_prefix`](https://github.com/geometalab/osmaxx/blob/c32f9dd8817db9ff6abf3027b4d028dc1c3d1adb/osmaxx/conversion/models.py#L66) which [slugifies the excerpt name](https://github.com/geometalab/osmaxx/blob/c32f9dd8817db9ff6abf3027b4d028dc1c3d1adb/osmaxx/conversion/models.py#L99) as well as some other components it puts into the string it returns.

hixi:
:+1:

PR #843, opened 2017-04-24
Feature/#820 allow accessing own existing exports without email address

### LINKED ISSUE #820: Allow accessing "my exports" & individual own exports without confirmed email address
There is no reason to disallow access to "my exports" for logged-in users without confirmed email address, as AFAIK no action available there needs the email address. Just being logged in should suffice.

das-g:
closes #820

### DIFF
--- a/osmaxx/excerptexport/views.py
+++ b/osmaxx/excerptexport/views.py
@@ -44,17 +44,13 @@ class OrderFormViewMixin(FormMixin):
         )
 
 
-class AccessRestrictedBaseView(LoginRequiredMixin, EmailRequiredMixin):
-    pass
-
-
-class OrderNewExcerptView(AccessRestrictedBaseView, OrderFormViewMixin, FormView):
+class OrderNewExcerptView(LoginRequiredMixin, EmailRequiredMixin, OrderFormViewMixin, FormView):
     template_name = 'excerptexport/templates/order_new_excerpt.html'
     form_class = ExcerptForm
 order_new_excerpt = OrderNewExcerptView.as_view()
 
 
-class OrderExistingExcerptView(AccessRestrictedBaseView, OrderFormViewMixin, FormView):
+class OrderExistingExcerptView(LoginRequiredMixin, EmailRequiredMixin, OrderFormViewMixin, FormView):
     template_name = 'excerptexport/templates/order_existing_excerpt.html'
     form_class = ExistingForm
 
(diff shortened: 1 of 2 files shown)


tests/excerptexport/views/test_export_list.py
```diff
@@ -41,21 +41,19 @@ def setUp(self):
             orderer=User.objects.create_user('another_user', 'someone_else@example.com', 'top secret')
         )
 
-    def test_permission_denied_if_lacking_email(self, *args):
+    def test_access_ok_even_without_confirmed_email(self, *args):
```
hixi: Should we add a test for testing without logged in user, right now we only test happy path - one could argue, that the logged in or not is being tested inside the Django framework itself, so no need to test it here.

osmaxx/excerptexport/views.py
```diff
@@ -44,17 +44,13 @@ def form_valid(self, form):
         )
 
 
-class AccessRestrictedBaseView(LoginRequiredMixin, EmailRequiredMixin):
-    pass
-
-
-class OrderNewExcerptView(AccessRestrictedBaseView, OrderFormViewMixin, FormView):
+class OrderNewExcerptView(LoginRequiredMixin, EmailRequiredMixin, OrderFormViewMixin, FormView):
```
hixi: Explicit is better then implicit :+1:

tests/excerptexport/views/test_export_list.py
das-g: I think it'd make sense to test here. Getting on it.

tests/excerptexport/views/test_export_list.py
```diff
@@ -41,21 +42,20 @@ def setUp(self):
...
-    def test_permission_denied_if_lacking_email(self, *args):
+    def test_redirect_to_login_if_not_logged_in(self, *args):
         response = self.client.get(
             reverse(
                 'excerptexport:export_list'
             )
         )
-        # redirect to 'Access Denied' page
-        self.assertEqual(response.status_code, 302)
+        self.assertRedirects(response, '/login/?next=/exports/', fetch_redirect_response=False)
```
das-g: By default, `assertRedirects` also tests whether fetching the target page gives a `200` status. Our tests settings are currently not configured so that our social login pages will work, thus I've skipped that step with `fetch_redirect_response=False`.

hixi: :+1: -> I think skipping is OK in this case.

Issue #780, opened 2017-01-11
saner sign-up/log-in/email-verification UX flow


das-g:
As a user, when I'm navigating somewhere where I can't get without doing something else first (log in, set an email address, etc.), I'd like to automatically get "back" to my initial destination once I've taken that hurdle. (Unless I explicitly navigate somewhere else by then.)

das-g:
This was closed by a wrong issue reference. (3871577507e51d480d5164519ded5129c4da84cd has now been replaced by a2b910d87a383cbbce827f1b89dca22085fa51cf.)

Re-opening.

PR #848, opened 2017-04-27
#780 return user to original page after login

das-g:
connected to #780 (part of it, actually)

This redirects the user to the page from where they came to our `/login` page, after they've logged in using an external identity/authentication provider (currently OpenStreetMap).

With the help of [the documentation](http://python-social-auth.readthedocs.io/en/latest/use_cases.html#return-the-user-to-the-original-page) this was surprisingly easy. Everything else was already in place. (The `next=` parameter when being redirected to our `/login` page is set by the `django.contrib.auth` machinery triggered by `LoginRequiredMixin`. The `@login_required` decorator on a function-style view would trigger a different mechanism to cause the same effect.)

### DIFF
--- a/osmaxx/core/templates/osmaxx/login.html
+++ b/osmaxx/core/templates/osmaxx/login.html
@@ -22,6 +22,6 @@
             {% endblocktrans %}
         </p>
 
-        <a style="font-size: 1.2em;" class="btn btn-default btn-success" href="{% url 'social:begin' 'openstreetmap' %}"><img src="{% static 'images/profile/osm_logo.svg' %}" alt="OSM Logo" height="32" width="32" />&nbsp;<strong>Log in</strong> using OpenStreetMap</a>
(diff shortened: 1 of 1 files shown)


osmaxx/core/templates/osmaxx/login.html
```diff
@@ -22,6 +22,6 @@
             {% endblocktrans %}
         </p>
 
-        <a style="font-size: 1.2em;" class="btn btn-default btn-success" href="{% url 'social:begin' 'openstreetmap' %}"><img src="{% static 'images/profile/osm_logo.svg' %}" alt="OSM Logo" height="32" width="32" />&nbsp;<strong>Log in</strong> using OpenStreetMap</a>
+        <a style="font-size: 1.2em;" class="btn btn-default btn-success" href="{% url 'social:begin' 'openstreetmap' %}?next={{ request.GET.next }}"><img src="{% static 'images/profile/osm_logo.svg' %}" alt="OSM Logo" height="32" width="32" />&nbsp;<strong>Log in</strong> using OpenStreetMap</a>
```
hixi: Suggestion: `{% if request.GET.next %}?next={{ request.GET.next }}{% endif %}`

PR #849, opened 2017-05-01
Feature/dont pass profile to send_email_verification

das-g:


osmaxx/profile/views.py
```diff
@@ -18,14 +18,15 @@
 class SendVerificationEmailMixin(object):
     RATE_LIMIT_SECONDS = 30
 
-    def _send_email_verification(self, profile):
-        if cache.get(profile.associated_user.id):
+    def _send_email_verification(self):
```
hixi: Do we ensure that the profile exists before calling this function? Before the caller had to pass the profile and take care of this. Maybe using an `assert user.profile` might remedy this issue?

das-g: The current callers ensure this or at least assume it, otherwise they wouldn't have been able to pass the profile. But we can't really guarantee it for future callers or future revisions of the current callers.

Would this be reason to keep the profile argument?

An alternative refactoring would be to make `_send_email_verification` a public method of `models.Profile`.

hixi: I think keeping the profile as parameter isn't a good idea, but having this method as public method on `Profile` sounds good to me.

das-g: > having this method as public method on Profile sounds good to me.

PR for that approach, as an alternative to this one: #850

hixi: I like that one much better, i puts readable names and clearer structure on the Profile module, which is the one that should be concerned with this, IMO.
I accidentally already merged it, if you don't agree with the change merge, I'll revert it - otherwise, I think we can close this issue.

das-g:
Closing in favor of #850

PR #850, opened 2017-05-01
Feature/move _send_email_verification to models.profile

das-g:
Alternative to #849 

While I like this functionality being accessed through the `Profile` instance, I felt like the implementation was too "view-y" and not enough "model-y" to belong into the model class, so I've made the method delegate to a free function in a new module.

I'm wondering that's good or whether a model mixin would be a better idea or whether it should just be put into the model proper regardless.

### DIFF
--- a/osmaxx/profile/email_confirmation.py
+++ b/osmaxx/profile/email_confirmation.py
@@ -7,13 +7,15 @@ from django.template.loader import render_to_string
 from django.urls import reverse
 from django.utils.translation import ugettext as _
 
+RATE_LIMIT_SECONDS = 30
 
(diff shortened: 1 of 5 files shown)


osmaxx/profile/email_confirmation.py
```diff
@@ -0,0 +1,45 @@
...
+from django.core.mail import send_mail
+from django.template.defaultfilters import urlencode
+from django.template.loader import render_to_string
+from django.urls import reverse
+from django.utils.translation import ugettext as _
+
+RATE_LIMIT_SECONDS = 30
+
+
+def send_email_confirmation(profile, request):
```
hixi: :+1:

Issue #771, opened 2016-12-07
Encoding problem in Mail notifications


LoremoCH:
Die Mail Benachrichtigungen von osmaxx haben ein Encoding Problem. Umlaute wie ä/ö/ü im Namen des Exportes (wie hier Süddeutschland) funktionieren hingegen. Getestet habe ich nur mit G-Mail (Web sowie Andorid App) und Thunderbird.

```
This is an automated email from osmaxx.hsr.ch

The extraction order #315 "Schweiz und Südeutschland" has been processed and is available for download:
- Garmin navigation &amp; map data: https://osmaxx.hsr.ch/media/osmaxx/outputfiles/somefile.blabla


View the complete order at https://osmaxx.hsr.ch/exports/ (login required)

Thank you for using OSMaxx.
The team at Geometa Lab HSR
geometalab@hsr.ch
```

das-g:
Ich nehme an, du meinst das `&amp;`?

LoremoCH:
Ja.

das-g:
While we could avoid this with an [`{% autoescape off %}`](https://docs.djangoproject.com/en/1.9/ref/templates/builtins/#autoescape) block in the affected template(s), that would be an IMHO semantically wrong approach, as it'd imply that we know the result is HTML-safe, while it in fact emphatically _isn't_;&mdash;we're just not using the rendered string in a plain text email thus why it doesn't have to be HTML-safe. Instead, I'd like to control the (not-)escaping from a part of the code where I _know_ about this usage, rather than in the template itself, which could at least theoretically also be used in an HTML context where escaping would be necessary.

Controlling this elegantly at render-time (which _is_ the one place where it can be seen how the resulting string will be further used in subsequent code) becomes possible in Django 1.10, by using a separately set up template engine that has the [`autoescape` option](https://docs.djangoproject.com/en/1.10/topics/templates/#module-django.template.backends.django) set to `False`.

Thus, let's update to Django 1.10 before tackling this.

das-g:
Status report:
* The update to Django 1.10 is done: #790
* The bugfix (WIP commit 8dda7f54cbc76e17f529b40675ada0a147b2d851) is held back for now, because it breaks django-crispy-forms. I've [re-opened Django issue \#27359](https://code.djangoproject.com/ticket/27359#comment:7) because of that and hope for replies that indicate how libraries can be fixed to work with django projects that configure multiple template engines of the same backend.

das-g:
[...]

PR #853, opened 2017-05-03
workaround #771: unescape HTML after rendering plain text email

das-g:
connected to #771 (not a final solution, just a hacky workaround)

The added test is the same as in WIP commit 8dda7f54cbc76e17f529b40675ada0a147b2d851 of the (hopefully) final solution that'll become available with Djagno 2.0

### DIFF
--- a/osmaxx/excerptexport/models/extraction_order.py
+++ b/osmaxx/excerptexport/models/extraction_order.py
@@ -3,6 +3,7 @@ from django.db import models
 from django.db.models.signals import post_save
 from django.dispatch.dispatcher import receiver
 from django.template.loader import render_to_string
+from django.utils.text import unescape_entities
 from django.utils.translation import ugettext_lazy as _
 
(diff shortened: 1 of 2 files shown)


osmaxx/excerptexport/models/extraction_order.py
```diff
@@ -86,10 +87,12 @@ def _get_all_exports_done_email_subject(self):
...
-        return render_to_string(
-            'excerptexport/email/all_exports_of_extraction_order_done_subject.txt',
-            context=view_context,
-        ).strip()
+        return unescape_entities(
+            render_to_string(
+                'excerptexport/email/all_exports_of_extraction_order_done_subject.txt',
+                context=view_context,
+            ).strip()
+        )  # HACK: calling unescape_entities as workaround for https://github.com/geometalab/osmaxx/issues/771
```
hixi: ugly, but it is what it is - I don't know of a better way to do this.

PR #858, opened 2017-05-04
use build-in os.scandir only

das-g:
as we don't support Python < 3.5 anymore


das-g:
> [...] as we don't support Python < 3.5 anymore

Actually, we do (for local development&mdash;I forgot). We also [test it on CI](https://travis-ci.org/geometalab/osmaxx/jobs/228774057#L478). Oops :flushed:.

Closing for now.

Issue #845, opened 2017-04-26
Refactor Structure and Code


hixi:
For good cleanup measure, we should go through another round of cleanup, roughly doing:

- [ ] x
- [ ] x
- [ ] x
- [ ] x

hixi:
Also see #847 and #846 for related refactorings.

PR #862, opened 2017-05-08
Feature/#845 conversion constants

das-g:
Rename and move `conversion_api` constants to `conversion/constants`. (And remove it from `INSTALLED_APPS`.)

connected to #845 (part of it)


osmaxx/conversion/constants/__init__.py
```diff
@@ -0,0 +1,6 @@
+"""
+Interface between OSMaxx frontend (Django-based web client) and the OSMaxx conversion service (mediator and worker).
+
+This package holds shared code. All OSMaxx components (frontend & conversion service) may depend on it.
+No code in this package may depend on other OSMaxx Django apps, especially not the component-specific ones.
+"""
```
das-g: Should this docstring be retained at all?

hixi: IMHO can be removed.

osmaxx/conversion/converters/converter.py
```diff
@@ -2,7 +2,7 @@
 from osmaxx.conversion.converters import converter_gis
 from osmaxx.conversion.converters import converter_pbf
 from osmaxx.conversion.job_dispatcher.rq_dispatcher import rq_enqueue_with_settings
-from osmaxx.conversion_api.formats import FGDB, SHAPEFILE, GPKG, SPATIALITE, GARMIN, PBF
+from osmaxx.conversion.constants.formats import FGDB, SHAPEFILE, GPKG, SPATIALITE, GARMIN, PBF
```
hixi: `from osmaxx.conversion.constants import formats` might be worth considering as an import and usage would become `formats.GARMIN`.

osmaxx/conversion/management/commands/result_harvester.py
```diff
@@ -12,7 +12,7 @@
 
 from osmaxx.conversion import models as conversion_models
 from osmaxx.conversion._settings import CONVERSION_SETTINGS
-from osmaxx.conversion_api.statuses import FINAL_STATUSES, FINISHED, FAILED
+from osmaxx.conversion.constants.statuses import FINAL_STATUSES, FINISHED, FAILED
```
hixi: Analogous to the above comment:
`from osmaxx.conversion.constants import statuses` might be worth considering as an import and usage would become `statuses.FINAL_STATUSES`.

osmaxx/conversion/converters/converter.py
das-g: Singular (`format.FGDB`) would make sense then, **but** there's already a [Python builtin called `format()`](https://docs.python.org/3/library/functions.html#format) that I don't want to shadow.

Any suggestions for alternative names?

hixi: We could use something along the lines of `options`, since having them capitalized already makes them constants by convention. But I think it makes more sense in this case to actually use a longer name: `output_format`.

PR #864, opened 2017-05-09
Feature/#845 conversion format

das-g:
Add and use convenience import for module `output_format`.

connected to #845 (kinda)

### DIFF
--- a/osmaxx/conversion/__init__.py
+++ b/osmaxx/conversion/__init__.py
@@ -1,1 +1,5 @@
+from .constants import output_format
+
 default_app_config = 'osmaxx.conversion.apps.ConversionConfig'
+
+__all__ = ['default_app_config', 'output_format']
--- a/osmaxx/conversion/constants/output_format.py
+++ b/osmaxx/conversion/constants/output_format.py
@@ -1,7 +1,7 @@
 import uuid
 from collections import OrderedDict, namedtuple
 
-from django.utils.translation import gettext as _
+from django.utils.translation import gettext_lazy as _
 
 FGDB, SHAPEFILE, GPKG, SPATIALITE, GARMIN, PBF = 'fgdb', 'shapefile', 'gpkg', 'spatialite', 'garmin', 'pbf'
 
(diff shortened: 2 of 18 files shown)


osmaxx/conversion/constants/output_format.py
```diff
@@ -1,7 +1,7 @@
 import uuid
 from collections import OrderedDict, namedtuple
 
-from django.utils.translation import gettext as _
+from django.utils.translation import gettext_lazy as _
```
hixi: Why do we need gettext_lazy here? Because lazy sounds better? :wink: 
http://stackoverflow.com/questions/4160770/when-should-i-use-ugettext-lazy

osmaxx/conversion/__init__.py
```diff
@@ -1 +1,5 @@
+from .constants import output_format
+
 default_app_config = 'osmaxx.conversion.apps.ConversionConfig'
+
+__all__ = ['default_app_config', 'output_format']
```
hixi: :+1:

tests/conversion/conftest.py
```diff
@@ -5,12 +5,12 @@
 import pytest
 from django.conf import settings
 
-import osmaxx.conversion.constants.output_format
+from osmaxx.conversion import output_format
```
hixi: :+1: for making it equal to the other imports

hixi:
I like this, it's only a cosmetic change, but makes the code much more readable :+1:

osmaxx/conversion/constants/output_format.py
das-g: Because with `gettext`, 34aa0e0 would lead to

    INTERNALERROR> django.core.exceptions.AppRegistryNotReady: The translation infrastructure cannot be initialized before the apps registry is ready. Check that you don't make non-lazy gettext calls at import time.

due to `output_format` [running `_(...)` at import time](https://github.com/geometalab/osmaxx/pull/864/files#diff-fc533de93246bf0b7f52b65e33f0ede5R45) and being imported all over the place.

(I'm not quite sure why e.g. the [import in `models.py`](https://github.com/geometalab/osmaxx/pull/864/files#diff-26f3feed2376939bed572ca895031e8fL11) didn't trigger this error even before.)

hixi: Forgot about the import in the models, this is correct (using gettext_lazy).

PR #873, opened 2017-07-12
fix #872: correct SRID for NAD 83

### LINKED ISSUE #872: NAD 83 should be EPSG:4269, not EPSG:4629
[EPSG:4**62**9 (Tahaa 54)](http://epsg.io/4629) ([SR-ORG](http://spatialreference.org/ref/epsg/4629/)) is for French Polynesia, while [EPSG:4**26**9 (NAD 1983)](http://epsg.io/4269) ([SR-ORG](http://spatialreference.org/ref/epsg/4269/)) is for the whole of North America (and AFAIK usable globally, even).

We currently have them mixed up and [transform to EPSG:4**62**9 when 'NAD 83' is selected](https://github.com/geometalab/osmaxx/blob/ab46256d5f60805858d6bf080e93eb384609ee73/osmaxx/conversion/constants/coordinate_reference_system.py#L8).

das-g:
Note that this doesn't change any existing data about past exports. Leaving them untouched should be appropriate as they would actually have been exported to EPSG:4629 (Tahaa 54), even though that would not have been the intention of the users.

fixes #872


osmaxx/conversion/constants/coordinate_reference_system.py
```diff
@@ -5,7 +5,7 @@
 WGS_84 = 4326
 WGS_72 = 4322
 PSEUDO_MERCATOR = 3857
-NAD_83 = 4629
+NAD_83 = 4269
```
hixi: this was _almost_ correct :see_no_evil:

das-g: Yeah, well, we hat it wrong on the [on the wiki](https://github.com/geometalab/osmaxx/wiki/Coordinate-Reference-Systems) from the very beginning and didn't properly check. :man_facepalming:

PR #879, opened 2017-09-06
fix syntax for docker-compose

hixi:


hixi:
Let's keep it as is. We now know what it would take. Closing.

PR #884, opened 2017-12-14
Feature/sane defaults

hixi:


Makefile
```diff
@@ -25,24 +25,12 @@ _list_targets_on_separate_lines:
...
-local_dev_env: compose-env/common.env compose-env/frontend.env compose-env/mediator.env compose-env/worker.env
+local_dev_env:
 	@echo
 	@echo "\tIf you haven't already, you might want to "'`source activate_local_development`', now.
 
-compose-env/common.env: compose-env-dist/common.env
-	@mkdir -p $(@D)
-	sed -e 's/^\(DJANGO_OSMAXX_CONVERSION_SERVICE_\(USERNAME\|PASSWORD\)=\)/\1dev/' < $< > $@
-
 PUBLIC_LOCALHOST_IP := $(shell ip route get 1 | awk '{print $$NF;exit}')
```
das-g: Seems like `PUBLIC_LOCALHOST_IP` isn't used anymore. If so, let's remove it, too.

hixi: I did find an issue, that `conversion_job.callback_url` is pointing to localhost:8888 which is probably caused by not using this setting anymore. Since this varies between developer machines, I'm not sure how we should handle this. I have a proposal I'll push shortly.

hixi:
Yes, that is a good proposal. Will push changes.

PR #891, opened 2018-05-16
add whitenoise middleware which got forgotten

hixi:
Added whitenoise middleware.


web_frontend/config/settings/production.py
```diff
@@ -7,6 +7,8 @@
 MIDDLEWARE_CLASSES = (
     # Make sure djangosecure.middleware.SecurityMiddleware is listed first
     'django.middleware.security.SecurityMiddleware',
+    # add whitenoise middleware
```
das-g: What is the added value of this comment? Instead, better comment on the required order (before all other middlewares, with the exception of SecurityMiddleware).

PR #901, opened 2019-05-27
Feature/update info

hixi:


Dockerfile.mediator
```diff
@@ -8,9 +9,10 @@ ENV DJANGO_OSMAXX_CONVERSION_SERVICE_USERNAME=default_user DJANGO_OSMAXX_CONVERS
 ENV NUM_WORKERS=5 DATABASE_HOST=mediatordatabase DATABASE_PORT=5432 APP_PORT=8901 APP_HOST=0.0.0.0
 
 MAINTAINER HSR Geometalab <geometalab@hsr.ch>
+ENV DEBIAN_FRONTEND=noninteractive
```
das-g: Should we reset this at the end of the Dockerfile? (I think it'll retain the value even into the running containers, if we don't. That might be unexpected when using apt in such a  container interactively, e.g. for debugging.)

Dockerfile.worker
```diff
@@ -1,143 +1,110 @@
-# this gdal image comes with support for FileGDB and has python3.4 already installed
-FROM geometalab/gdal-debian:latest
+# this gdal image comes with support for FileGDB and has python3.6 already installed
+# is derived from ubuntu
```
das-g: This makes the comment non-grammatical.

Dockerfile.worker
```diff
@@ -1,143 +1,110 @@
...
+FROM geometalab/gdal-docker:v3.0.0
 
 USER root
 
-ENV PYTHONUNBUFFERED=non-empty-string PYTHONIOENCODING=utf-8 LC_ALL=C.UTF-8 LANG=C.UTF-8
+ENV PYTHONUNBUFFERED=non-empty-string PYTHONIOENCODING=utf-8 LC_ALL=C.UTF-8 LANG=C.UTF-8 DEBIAN_FRONTEND=noninteractive
```
das-g: Should we reset `DEBIAN_FRONTEND` at the end of the Dockerfile? (I think it'll retain the value even into the running containers, if we don't. That might be unexpected when using apt in such a container interactively, e.g. for debugging.)

docs/development/project-development-environment.md
```diff
@@ -133,7 +133,7 @@ ip route get 1 | awk '{print $NF;exit}'
 
 You can generate the complete URL in `sh` with:
 ```bash
-echo "http://$(ip route get 1 | awk '{print $NF;exit}'):8888"
+echo "http://$(ip route get 1 | awk '{print $NF;exit}'):8889"
```
das-g: Müsste das `awk`-Kommando hier auch angepasst werden?

tests/conftest.py
```diff
@@ -205,6 +207,24 @@ def user(db, django_user_model, django_username_field):
     return user
 
 
+def create_authenticated_client(client, user):
```
das-g: fixtures can AFAIR also be called like normal functions, so why not just inline this in `authenticated_client` below?

Dockerfile.frontend
```diff
@@ -1,5 +1,6 @@
-# this gdal image comes with support for FileGDB and has python3.4 already installed
-FROM geometalab/gdal-debian:latest
+# this gdal image comes with support for FileGDB and has python3.6 already installed
+# is dervied from ubuntu
```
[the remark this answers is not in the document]
hixi: I read them as two separate comments, has spelling errors anyway.

docs/development/project-development-environment.md
hixi: Yes, missed that one, thanks.

tests/conftest.py
hixi: Nope, not anymore. That behaviour changed in pytest versions, don't know where exactly; that's actually the only reason I changed this.

Dockerfile.mediator
hixi: Good point.

Dockerfile.frontend
```diff
@@ -1,13 +1,17 @@
-# this gdal image comes with support for FileGDB and has python3.4 already installed
-FROM geometalab/gdal-debian:latest
+# This gdal image comes with support for FileGDB and has python3.6 already installed.
```
das-g:
```suggestion
# This GDAL image comes with support for FileGDB and has Python 3.6 already installed.
```

Dockerfile.frontend
```diff
@@ -1,13 +1,17 @@
...
+# Based ubuntu.
```
das-g:
```suggestion
# Based on image _/ubuntu.
```
or
```suggestion
# Based on official Ubuntu docker image.
```

Dockerfile.mediator
```diff
@@ -1,5 +1,7 @@
-# this gdal image comes with support for FileGDB and has python3.4 already installed
-FROM geometalab/gdal-debian:latest
+# This gdal image comes with support for FileGDB and has python3.6 already installed.
```
das-g:
```suggestion
# This GDAL image comes with support for FileGDB and has Python 3.6 already installed.
```

Dockerfile.mediator
```diff
@@ -1,5 +1,7 @@
...
+# Based ubuntu.
```
das-g:
```suggestion
# Based on image _/ubuntu.
```
or
```suggestion
# Based on official Ubuntu docker image.
```

Dockerfile.worker
```diff
@@ -1,143 +1,114 @@
-# this gdal image comes with support for FileGDB and has python3.4 already installed
-FROM geometalab/gdal-debian:latest
+# This gdal image comes with support for FileGDB and has python3.6 already installed.
```
das-g:
```suggestion
# This GDAL image comes with support for FileGDB and has Python 3.6 already installed.
```

Dockerfile.worker
```diff
@@ -1,143 +1,114 @@
...
+# Based ubuntu.
```
das-g:
```suggestion
# Based on image _/ubuntu.
```
or
```suggestion
# Based on official Ubuntu docker image.
```

